PR #292, opened 2015-01-13
Add annotations configuration in generator

showaltb:
For Rails 4.0+, the generator will add configuration entries to
config/environments/development.rb to include Cucumber's features
directory and .feature files in the "rake notes" run.

### DIFF
--- a/features/annotations.feature
+++ b/features/annotations.feature
@@ -1,0 +1,23 @@
+@requires-rails-version-42
+Feature: Annotations
+
+  I order to track my development progress
+  As a developer
+  I should be able to list annotations in my features
+
+  Background:
+    Given I have created a new Rails app "test-app" with no database and installed cucumber-rails
+    And I write to "features/tests.feature" with:
+      """
+      Feature: Tests
+        Scenario: Tests
+          # TODO: When I go to the products page
+      """
+
+  Scenario: See annotations in .feature file
+    When I run `bundle exec rake notes`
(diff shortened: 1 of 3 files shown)


showaltb:
The Travis CI build is failing, but I cannot see how the failure has anything to do with my changes?

Kosmas:
@showaltb thank you. I'm going to have a look too. Do the tests pass on your local?

showaltb:
@Kosmas, no, the tests do not pass on my local, even for master branch. The error is some kind of timeout, but it may be the same thing Travis CI is seeing. Unfortunately, I don't have any time now to dig deeper, sorry. The specs that I added to pass both on local and on Travis CI.

Kosmas:
@showaltb would you like to try again now that the tests are fixed?

showaltb:
I rebased this against current master, and it builds, but don't merge it yet.

It doesn't work against latest rails (4.2.5), because `rake notes` doesn't load the environment, so the code in `config/environments/development.rb` doesn't run. Running `rake environment notes` works.

I need to do more investigation here.

showaltb:
OK, this is good to merge now. I have re-written it to configure the annotations through the rake task, similar to how you are configuring the stats task. I also added a feature to test the new functionality. The new functionality is only enabled on Rails 4.2+.

Kosmas:
Thank you @showaltb !

PR #294, opened 2015-03-30
Depend on railties instead of rails

langalex:
...so that projects that for example don't need ActiveRecord don't automatically have to include it (rails depends on activerecord etc., railties does not).

### DIFF
--- a/cucumber-rails.gemspec
+++ b/cucumber-rails.gemspec
@@ -15,7 +15,8 @@ Gem::Specification.new do |s|
   s.add_runtime_dependency('capybara', ['>= 1.1.2', '< 3'])
   s.add_runtime_dependency('cucumber', ['>= 1.3.8', '< 2.1'])
   s.add_runtime_dependency('nokogiri', '~> 1.5')
-  s.add_runtime_dependency('rails', ['>= 3', '< 5'])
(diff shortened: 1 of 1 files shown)


langalex:
no idea why rails3 is failing. unfortunately it works on my box. any pointers?

Kosmas:
@langalex that is because running the tests on CI, involves using Appraisal.
And Appraisal creates different gemfiles for the different rails versions.

https://github.com/cucumber/cucumber-rails/tree/master/gemfiles

So you would need to replace the rails with railties in these gemfiles too..
Could you try and replace them and try again?

By the way you can see why is failing when you go to Travis and click on the job number (ie #280.7).
It sounds like a good idea, but:

a) I believe that the most common case of using inside rails is with ActiveRecord?
b) Should we also rename it to cucumber-railties ? :;-)

langalex:
@Kosmas I had to try a few combinations and in the end, adding rails as a development dependency was the only one that worked.

a) yes, that is the most common case, but there are quite a few people not using AR, plus this change has zero impact on the default case, so I think there's no harm in making the change
b) I'll leave that up to you :P

Kosmas:
Thank you!

PR #296, opened 2015-06-03
Disable display of detailed exceptions

pedrosmmoreira:
`@allow-rescue` permits rescuing exceptions by setting `action_dispatch.show_exceptions` to true in the env hash. However, this will render the detailed error page and not the public error page you'd encounter in production.

Given that the feature enables testing of error pages, and seems to be intended to mirror production behaviour as per the comment on https://github.com/cucumber/cucumber-rails/blob/master/lib/generators/cucumber/install/templates/support/_rails_each_run.rb.erb#L6-L14, this PR sets `action_dispatch.show_detailed_exceptions` to false, thus enabling the public error page to be rendered.

### DIFF
--- a/features/allow_rescue.feature
+++ b/features/allow_rescue.feature
@@ -32,8 +32,7 @@ Feature: Allow Cucumber to rescue exceptions
         visit '/posts'
       end
       Then /^I should see the public error page$/ do
-        expect(page).to have_content "We're sorry, but something went wrong. \
-          If you are the application owner check the logs for more information."
+        expect(page).to have_content "We're sorry, but something went wrong."
       end
       """
     And I run `bundle exec rake db:migrate`
--- a/features/allow_rescue.feature
+++ b/features/allow_rescue.feature
@@ -23,12 +23,17 @@ Feature: Allow Cucumber to rescue exceptions
       Feature: posts
         @allow-rescue
         Scenario: See posts
-          When I look at the posts
+        When I look at the posts
+        Then I should see the public error page
       """
     And I write to "features/step_definitions/posts_steps.rb" with:
       """
       When /^I look at the posts$/ do
         visit '/posts'
+      end
+      Then /^I should see the public error page$/ do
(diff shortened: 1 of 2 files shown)


Kosmas:
@pedrosmmoreira thank you for the PR, but the tests are not passing. Can you please try and get them to pass first?

pedrosmmoreira:
@Kosmas Absolutely, I'm working on it :)

pedrosmmoreira:
@Kosmas This PR depends on `action_dispatch.show_detailed_exceptions` being set, which was only introduced in Rails 3.2.1, as far as I can tell. Hence, the appraisal for previous versions is failing and breaking the build. How would you like me to proceed?

Kosmas:
@pedrosmmoreira thanks for having a look.

I think it would make sense to have a check for the rails version, and then depending on > or < 3.2.1 set the show_detailed_exceptions. What do you think?

Kosmas:
@pedrosmmoreira would you like to try again now that the tests are fixed?

pedrosmmoreira:
@Kosmas Sure, I'll rebase and check if we can clear this.

Kosmas:
@pedrosmmoreira can you please rebase so it can run again.
It should be working now.

pedrosmmoreira:
@Kosmas Done. Sorry for the delay on this.

Kosmas:
@pedrosmmoreira no worries.. we had some problems with the tests anyway, but let's see if they work now!

mathieujobin:
@pedrosmmoreira can you rebase with master to make sure all tests passes?
thanks

pedrosmmoreira:
@mathieujobin All done - I'm currently traveling but I'll look to see if we can this ready for merge as soon as I am back. Apologies for leaving it neglected so long!

xtrasimplicity:
I've rebased these changes onto master, and have created a new PR ([#387](https://github.com/cucumber/cucumber-rails/pull/387)). I'll close this one off.

Issue #299, opened 2015-09-02
Tests fail on clean clone of gem


rstawarz:
Cloning this gem cleanly (i.e. a new clone) fails to run tests due to using a newer version of [`cucumber-aruba`](https://github.com/cucumber/aruba/releases/tag/v0.9.0).  Turns out `> 0.8.0` cause the console tests to fail due to incompatibilities.

PR incoming for a fix to lock the gem.

PR #300, opened 2015-09-02
Lock cucumber-aruba gem to `< 0.8`

rstawarz:
The cucumber-aruba gem has had several releases.  Starting from a clean
clone of cucumber-rails, `bundle install` & rake fail since the latest
version of the gem is not compatible.

Fixes issue #299

### DIFF
--- a/cucumber-rails.gemspec
+++ b/cucumber-rails.gemspec
@@ -21,7 +21,7 @@ Gem::Specification.new do |s|
   # Main development dependencies
   s.add_development_dependency('ammeter', ['>= 0.2.9', '< 2'])
   s.add_development_dependency('appraisal', '>= 0.5.1')
-  s.add_development_dependency('aruba', '>= 0.4.11')
+  s.add_development_dependency('aruba', ['>= 0.4.11', '< 0.8'])
   s.add_development_dependency('builder', ['>= 2.1.2', '< 4'])
   s.add_development_dependency('bundler', '>= 1.3.5')
   s.add_development_dependency('selenium-webdriver', '>= 2.45.0')


Kosmas:
@rstawarz would you like to try again now that the tests are fixed?

mvz:
I think for Rails 3.0, the current Gemfile results in 0.8.1 being installed. It's probably best to lock to that version.

DawidJanczak:
I checked this on my local box and `<0.8` won't work because of undefined method `set_environment_variable` That being said @rstawarz is right in that just cloning the repo and running `rake` after installing gems fails. Without changes to `gemspec` aruba `0.10.2` is installed, but the failure is the same on `0.8.0`, namely `bundle exec rails generate cucumber:install` fails in here: https://github.com/cucumber/cucumber-rails/blob/master/features/step_definitions/cucumber_rails_steps.rb#L22 No error is printed, but I think the issue here is that gems are added to Gemfile, but `bundle install` is not run and `bundle exec` fails because of that.
Please do let me know if I'm missing a step or two.

Kosmas:
thanks @tooky any idea about the failing tests? (cucumber features/rest_api.feature:3 # Scenario: Compare JSON)

tooky:
@Kosmas mea culpa - hadn't tested enough versions before I pushed :blush:

I think the issue is related to aruba's changing timeout configuration. I've pushed a fix.

Kosmas:
@tooky no problem. I was about to ask for your help as there were two failing tests here too: https://travis-ci.org/Kosmas/cucumber-rails that were not letting me release a new version.
It would be great if you can find out the problem!

tooky:
@Kosmas master build is fixed now - worth rebasing to see if it fixes your fork?

Kosmas:
@tooky that is great :+1: 
thank you everything works now.
will try and do a release tomorrow!

PR #303, opened 2015-11-03
Get started fixing tests

wied03:
Does not fix all the features, but it's a start


Kosmas:
@wied03 that looks promising. Thanks. I was trying going the same route changing the order of the requires in the spec_helper.rb but did not go far. There are a couple of them that fail because of the capybara version. Do you want to give it another try before doing a PR, please?

wied03:
Yeah, I realized I hadn't touched any of those gemfiles (was too busy hacking away @ the main one trying to see what changed and broke this).

mattwynne:
Fixed by d5871533af32a6b1d2f7a1a02a65cc128b2c93ae

PR #304, opened 2015-11-26
Expand support to include mime-types 3

halostatue:
The tests pass locally. Built on top of #303 because the tests are failing
   for reasons other than mime-types changes.

### DIFF
--- a/cucumber-rails.gemspec
+++ b/cucumber-rails.gemspec
@@ -16,7 +16,7 @@ Gem::Specification.new do |s|
   s.add_runtime_dependency('cucumber', ['>= 1.3.8', '< 3'])
   s.add_runtime_dependency('nokogiri', '~> 1.5')
   s.add_runtime_dependency('railties', ['>= 3', '< 5'])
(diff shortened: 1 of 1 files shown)


halostatue:
Please note that only 22fc0b1 is needed for this particular change, but that I am building on top of #303 by @wied03 so that the has a better chance of passing (my local tests were failing for reasons I neither have time nor interest in verifying).

Given that there’s no code in cucumber-rails that appears to call anything in mime-types, I’m not sure why this dependency exists, but we want to get people moved to mime-types 3 sooner, rather than later, as long as they are using Ruby 2.0 or higher.

Kosmas:
@halostatue would you like to try again now that the tests are fixed?

halostatue:
I will rebase and try again.

halostatue:
Yay.  This passes remotely. I am seeing this locally when running the features:

```
undefined method `set_environment_variable' for #<Object:0x007fb8c22c3f58> (NoMethodError)
```

I really don’t know what’s going on there causing this.

Kosmas:
thanks @halostatue

PR #311, opened 2016-03-30
Depend on railties instead of rails

clupprich:
As far as I can see there is no need to depend on `rails` as a whole, so `railties` only should be enough. That has the benefit that people who are not using e.g. `activerecord` don't get that as an extra dependency when including `cucumber-rails`.


clupprich:
Hm, that was unexpected. Do you have any tips how I can debug that? It seems to fail when running `run_simple 'bundle exec rails generate cucumber:install'`. It does so also when I checkout `master`, so I'm probably missing some dependency?

clupprich:
Disregard this PR, please. Reading it again, I can clearly see `add_runtime_dependency` vs `add_development_dependency` ;)

Kosmas:
No problem @clupprich !

PR #314, opened 2016-04-19
Copy cucumer executable to bin folder in Rails 4

zillou:
Rails introduced `bin` folder for executables and removed `script` folder since Rails 4. https://github.com/rails/rails/pull/8786

This PR will detect the rails version,  and copy the cucumber executable to the right place.


mathieujobin:
do we need this executable at all ? I find that everything works fine without it ..

zillou:
@mathieujobin if we don't need the `cucumber` executable anymore, can we just change the generator not to create a `script` folder and copy the `cucumber` script?

PR #326, opened 2016-09-27
Try cucumber 3.0.0.pre.1

tooky:
Creating this PR to check any issues arising with Cucumber 3.0.0.pre.1.


Kosmas:
@tooky looks good. 
OK to merge?

tooky:
@Kosmas I'm going to make a commit bumping the allowed version in the gemspec on master, incrementing the version of cucumber-rails, and reference this build there. I don't think we need to add the appraisals specifically for this to master? We can then release this in advance of cucumber 3.0, and it will allow people to use 3.0.0.pre.1 in their rails projects. WDYT?

Kosmas:
@tooky that is fine. 
Let me know when you want me to merge and then release the 1.4.5.

Kosmas:
@tooky version 1.4.5 with your changes released.

Issue #333, opened 2016-11-30
Drop support for older Rails versions?


mvz:
I've been trying to update the development dependencies to a more recent version of Aruba, and I found this is very hard to do in a clean way, since dependencies of Rails 3.0 and 3.1 indirectly force the use of Aruba 0.8.

Since Rails 3.0 and 3.1, and even 3.2, are ancient and unsupported, I would propose dropping support for it. People stuck on those platforms can simply use cucumber-rails 1.45 or earlier.

This would also allow for cleaning up of the generation of `script/cucumber` (#277), which can be replaced by a recommendation of the `spring-commands-cucumber` gem.

WDYT?

Kosmas:
@mvz I think that is a good idea and it makes sense.
Please go ahead!

mvz:
The implementation of this is ready in #334.

Kosmas:
Thank you @mvz

PR #334, opened 2016-12-01
Drop Rails 3 support

mvz:
Implements #333.

Also:
* Adds an appraisal to specifically test compatibility with the oldest declared versions of dependencies. 
* Updates the development dependencies to require compatible versions
* Updates aruba to ensure cucumber-rails' own scenarios see the correct bundle
* Restores compatibility of cucumber-rails' scenarios with cucumber 1.3.x


Kosmas:
thank you @mvz for all the hard work :+1:

PR #337, opened 2017-04-28
Support Rails 5.1

rafamanzo:
With the release of 5.1 updating from 5.1.0.rc2 was a problem due to cucumber-rails' railties restriction.

This adds a new appraisal for 5.1 and the necessary fix to the installation process on the acceptance test caused by duplicated capybara occurences on Gemfile since this new rails version already includes it.


mvz:
Nearly all the tests are failing on Travis but can't believe that's due to these changes.

mvz:
It seems we need to ensure rake < 11.0 is used on some of the appraisals: https://stackoverflow.com/questions/35893584/nomethoderror-undefined-method-last-comment-after-upgrading-to-rake-11#35893625

rafamanzo:
Thank you for digging this up @mvz!

I have pushed some new code with what you've suggested. Most of the builds should get fixed, but I think the 5.1 build will still fail. Do you have any other suggestions regarding this?

mvz:
It looks like putting the right rake version in the appraisals wasn't enough, because the Travis build environment still contains rake 12, and that's the one that gets picked up by the Gemfile that's created inside the cucumber scenarios. The nicest way to fix this is probably to add the correct version of rake explicitly in `features/step_definitions/cucumber_rails_steps.rb`, in the `#install_cucumber_rails` method.

rafamanzo:
Still no luck :/

mvz:
I've made some progress in #338.

mvz:
@rafamanzo I've fixed the build in #338. It's basically your two commits plus two more. Feel free to cherry pick my commits as needed.

rafamanzo:
Thanks @mvz! I've cherry-picked and as soon as #338 gets accepted I will rebase to clean up this PR.

I was able to fix one more test case, but I'm still fighting selenium-webdriver for the final 4 remaining failures. I've tried to update it to 3.4.0 locally with no success on getting these to pass.

mvz:
Ok, right, since not all tests pass in this PR, it does make more sense for #338 to be merged first.

Kosmas:
@mvz this seems OK now thanks to @rafamanzo .

mvz:
Merged it is! Thanks @rafamanzo.

PR #341, opened 2017-05-12
Update .travis.yml

junaruga:
* Updated Rubies to latest version.
* Added ruby-head as allow_failures.
  Because it's good to know new version Ruby's issue as faster before the release.
* Added ruby 2.4.1 as allow_failures to share the test failure.
  => Let me try to test 2.4.1 on Travis.

The motivation is same with https://github.com/cucumber/cucumber-ruby/pull/1087 .

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,8 +1,10 @@
 rvm:
   - 2.0.0
-  - 2.1.9
-  - 2.2.5
-  - 2.3.1
+  - 2.1.10
+  - 2.2.7
+  - 2.3.4
+  - 2.4.1
+  - ruby-head
 
 # whitelist
 branches:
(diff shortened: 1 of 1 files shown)


junaruga:
It looks Rails 4.1 have not supported Ruby 2.4. So, I will exclude the case.
https://github.com/rails/rails/blob/4-1-stable/.travis.yml

Rails 4.2 latest version have supported Ruby 2.4.
https://github.com/rails/rails/blob/4-2-stable/.travis.yml

Kosmas:
@junaruga thanks.

junaruga:
@Kosmas If you like, I will be able to update `History.md` for this PR.
When I checked the file yesterday, I thought that only important features and bugs were updated in case of `cucumber-rails`.

junaruga:
@Kosmas you have already updated it . Thank you for that!

PR #348, opened 2017-10-27
Conform the new syntax for tag expression

mirelon:

PR #359, opened 2018-02-12
Suppress deprecation warnings but maintain support for older Ruby versions (tag negation)

xtrasimplicity:
Following on from https://github.com/cucumber/cucumber-rails/pull/348


xtrasimplicity:
Closed in favour of https://github.com/cucumber/cucumber-rails/pull/360 - if successful, we can then restore @mirelon's changes completely (i.e. remove complete support for ~ tag negation).

PR #364, opened 2018-02-13
Re-adding tests for Rails 4.0

xtrasimplicity:
Just wanted to have CI run through this and make sure Rails 4 is still compatible. If not, we can update the gemspec file to cease support for Rails 4.0, too.


xtrasimplicity:
Well, that was a flop ([Ruby 2.2.7](https://travis-ci.org/cucumber/cucumber-rails/jobs/340861941), [Ruby 2.3.4](https://travis-ci.org/cucumber/cucumber-rails/jobs/340861946))! Looks like we'll have to reintroduce the `lowest_version_bounds` appraisal and test out Rails 4.0 on some older versions of RSpec. Happy for someone else to take over if they'd like - otherwise, I'll try to take another look at this sometime over the next few days.

PR #368, opened 2018-03-14
Update for ammeter 1.1.4

mvz:
The latest version of ammeter loads capybara/rails, which in turn means Rails.application needs to be present. Somewhere, a connection pool is needed too.


mvz:
:tada:

PR #376, opened 2018-04-10
[WIP] Support Rails 5.2.X

roberts1000:
## Summary

Allow cucumber-rails to bundle with projects that use Rails 5.2.X.

## Types of changes

## Checklist:


roberts1000:
@xtrasimplicity @mvz I'm getting the following error

```
LoadError: cannot load such file -- bootsnap/setup
```

I'm not really sure why the test Rails apps can't find bootsnap.  It appears to be in their Gemfile.  Do you have any suggestions on how to get past that issue?

xtrasimplicity:
@roberts1000 The gems defined in the appraisal should be loaded when the temporary rails app starts, but if not you can force a gem to be added to the temporary application's gem file from within the step definitions.

The following should give you an idea of how we've handled this in the past. :) 

```ruby
def install_cucumber_rails(*options)
    [...] 
    # From Rails 5.1 some gems are already part of the Gemfile
    if Gem.loaded_specs['rails'].version < Gem::Version.new('5.1.0')
      gem 'capybara', group: :test
      gem 'selenium-webdriver', group: :test
    end
  [...] 
end
```
Source: [features/step_definitions/cucumber_rails_steps.rb](https://github.com/cucumber/cucumber-rails/blob/master/features/step_definitions/cucumber_rails_steps.rb)

Apologies for any strange formatting - I'm using my phone.

Hope this helps. :)

roberts1000:
Thanks for the suggestion!  I gave it a shot but it doesn't seem to work with the updated Appraisal files or by modifying `install_cucumber_rails`.  I wonder if there's something specific to bootsnap that is causing problems.

roberts1000:
It has `gem 'bootsnap', '>= 1.1.0', require: false`.  The temporary rails app is correctly using `gem 'rails', '~> 5.2.0` (although the app folder is named `rails-3-app`).  If I `cd` into the /home/.../cucumber-rails/tmp/aruba/rails-3-app`  folder and run `bundle`, the temp app bundles fine.  I can also start the rails server and send requests to it without a problem.

xtrasimplicity:
@roberts1000 What is printed to stdout when you use the `@announce` tag? Could you please copy and paste it into a Gist?

https://relishapp.com/cucumber/aruba/v/0-9-0/docs/announce-output-during-test-run

xtrasimplicity:
I'm going to close this, as it is covered by https://github.com/cucumber/cucumber-rails/pull/378 - which will soon be merged. Thanks for all of your effort, @roberts1000!

roberts1000:
Thanks for helping with this PR and getting the problem solved in the other PR!  Wish I could have done more, but I got sidetracked by Rails Conf and travel.

PR #378, opened 2018-04-12
[WIP] Add Rails 5.2 & Capybara 3 support.

gobijan:
## Summary

Relax the runtime dependencies to allow railties version 5.2.

## Details

Relax the runtime dependencies to allow railties version 5.2.

## Motivation and Context

Allow Rails 5.2 now that it's released.

## How Has This Been Tested?

No introduction of new tests. Relying on tests that are already there.

## Screenshots (if appropriate):
none.

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue).
- [ ] New feature (non-breaking change which adds functionality).
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected).

## Checklist:

- [ ] I've added tests for my code.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.

PR #386, opened 2018-04-21
factory_girl renamed to factory_bot

varyonic:
Clean up deprecation warnings resulting from [Thoughtbot's decision to rename factory_girl](https://robots.thoughtbot.com/factory_bot).


varyonic:
Wow, was not watching #378. Thanks!

PR #391, opened 2018-11-20
Add support for Rails edge

jacobbednarz:
We're in the process of getting onto Rails edge and this gem is one of
our dependencies that needs to be updated to allow running rails > 5.

I've opted for "rails-edge" over 6 here as can be left as is once rails
6 is released and people will be able to continue upgrading to the next
version with less friction. I can be swayed to do "rails-6" if you'd
prefer though :)


jacobbednarz:
There is a question of which MRI versions this project wants to continue going forward as well. Rails edge does require 2.4.1+ which would get rid of 2.2 and 2.3 from the test suite. 2.2 is already EOL and 2.3 is pretty close to it but I'm not sure on your feelings in this area.

CI does have these failures but we can drop this in another PR if you'd like to go that route.

gemfiles/rails_5_2.gemfile
```diff
@@ -4,14 +4,14 @@ source "https://rubygems.org"
 
 gem "rails", "~> 5.2"
 gem "railties", "~> 5.2"
-gem "capybara", "~> 3.0"
+gem "capybara", "~> 3"
```
luke-hill: This actually does nothing, and is kind of a bug, so we should just leave this as is (The new styling would allow any version of capybara)

gemfiles/rails_5_2.gemfile
```diff
@@ -4,14 +4,14 @@ source "https://rubygems.org"
...
 gem "selenium-webdriver", "~> 3.11"
-gem "geckodriver-helper", "~> 0.0.5"
+gem "geckodriver-helper", "~> 0.20.2"
 gem "rspec-rails", "~> 3.7"
 gem "sass-rails", "~> 5.0"
-gem "coffee-rails", "~> 4.2"
+gem "coffee-rails", "~> 4.2.0"
```
luke-hill: Is this change and the one below intended here? If so they're a bit varied, so doesn't quite make sense

Appraisals
```diff
@@ -94,3 +94,16 @@ appraise "rails_5_2" do
...
+  gem "capybara", "~> 3"
+  gem "selenium-webdriver", "~> 3.11"
+  gem "geckodriver-helper", "~> 0.20.2"
+  gem "rspec-rails", "~> 3.7"
+  gem "sass-rails", "~> 5.0"
+  gem "coffee-rails", "~> 4.2.0"
+  gem "uglifier", "~> 4.1"
+  gem "sqlite3", "~> 1.3.13"
+  gem "database_cleaner", ">= 1.1"
+end
```
luke-hill: could you add a new line here

gemfiles/rails_5_2.gemfile
jacobbednarz: running `appraisal install` did this and I made the assumption it knew what it was doing. will remove though.

jacobbednarz:
Feedback addressed in fbb41aa and 6b1a0a4 respectively.

Appraisals
```diff
@@ -94,3 +94,17 @@ appraise "rails_5_2" do
   gem "sqlite3", "~> 1.3.13"
   gem "database_cleaner", ">= 1.1"
 end
+
+appraise "rails_edge" do
+  gem "rails", git: "https://github.com/rails/rails.git"
+  gem "capybara", "~> 3"
```
olleolleolle:
```suggestion
  gem "capybara", "~> 3.0"
```

Appraisals
```diff
@@ -94,3 +94,17 @@ appraise "rails_5_2" do
...
+  gem "selenium-webdriver", "~> 3.11"
+  gem "geckodriver-helper", "~> 0.20.2"
+  gem "rspec-rails", "~> 3.7"
+  gem "sass-rails", "~> 5.0"
+  gem "coffee-rails", "~> 4.2.0"
```
olleolleolle:
```suggestion
  gem "coffee-rails", "~> 4.2"
```

luke-hill:
Closing in favour of #405

PR #394, opened 2018-11-29
Get CI green

deivid-rodriguez:
## Summary

Just trying to get CI consistently green before proposing other changes.

## Details

There were two errors happening randomly but very often:

* A [timeout error](https://travis-ci.org/cucumber/cucumber-rails/jobs/461252725) error. Increasing selenium's timeout seems to fix it.

* [A problem](https://travis-ci.org/cucumber/cucumber-rails/jobs/459973208) with `geckodriver`. This seems like a problem with the `geckodriver-helper` gem, probably something equivalent to https://github.com/flavorjones/chromedriver-helper/issues/57. Since the gem was doing very little (download geckodriver and putting it in PATH), I decided to remove it.

## Motivation and Context

Get CI green to unblock other PRs.

## How Has This Been Tested?

My fork on TravisCI.

## Types of changes

## Checklist:


deivid-rodriguez:
Yeah, the only downside is that contributors now need to manually put `geckodriver` on their path. I added the install script to the contributing instructions so it's not too painful.

There's a separate alternative which does not have that issue: using the [webdrivers](https://github.com/titusfortner/webdrivers/) gem. I think this gem should not suffer from the same issues present before this PR because it does not put any binary on the user's path (and thus doesn't mess with rubygems binstubs) but instead downloads it and then configures selenium to use it.

I still think this solution is simple enough. Contributors similarly require `firefox` installed on their system for specs to work, and that's not solved by any gem. I think it's fine that external non ruby dependencies are setup manually.

features/step_definitions/cucumber_rails_steps.rb
```diff
@@ -98,9 +97,12 @@ def fixture(path)
 Given /^I force selenium to run Firefox in headless mode$/ do
   selenium_config = %{
     Capybara.register_driver :selenium do |app|
+      http_client = Selenium::WebDriver::Remote::Http::Default.new
```
luke-hill: This isn't directly needed. But to hazard a guess is this extra timeout something that causes intermittent failures>?

deivid-rodriguez: Read the PR description! :)

deivid-rodriguez:
I was about to add the `webdrivers` gem, but it also feels nice about this solution to consistently use a specific version of `geckodriver` instead of using always the latest. May I leave it as it is, and maybe iterate in the future if we end up considering the `webdrivers` gem a better approach?

xtrasimplicity:
Let's go for it. :)

aslakhellesoy:
Hi @deivid-rodriguez,

Aslak Hellesøy
Creator of Cucumber

PR #395, opened 2018-11-29
Simplify spec dependencies

deivid-rodriguez:
## Summary

While investigating why CI failed, I noticed that the dependencies are much more complicated that needed... unless I'm missing something.

## Details

This PR centralizes the dependencies in a single place. Most of them are only needed in the Gemfile of the generated rails application where cucumber is run, so they are only specified there. In order to do that, I added support for specifying versions in the [`gem` helper method](https://github.com/cucumber/cucumber-rails/pull/395/commits/e3f4dbe4a9b9d5fc92f4195d39705dda4142464e).

It also includes some other minor warning fixes and cleanups, see the specific commits for more information.

## Motivation and Context

I think this should simplify how dependencies are managed going forward.

## How Has This Been Tested?

By running the specs and checking that they pass.

## Types of changes

## Checklist:


deivid-rodriguez:
I rebased this PR after #396 was merged and should be now ready for review. Sorry if it's too big, I can split it if needed.

deivid-rodriguez:
Thanks @olleolleolle! Getting this in, then!

PR #408, opened 2019-04-10
Update development setup instructions

mvz:
## Summary

Update instructions to make sure all development dependencies are installed.

## Details

Mention the `install_geckodriver.sh` and `install_webpacker.sh` in both README and CONTRIBUTING.

## Motivation and Context

I completely missed `install_geckodriver.sh` because I looked at the README. So others might miss it too.

## Types of changes

* Documentation


mvz:
> Makes me think: the "GitHub standard" way of using bin/setup to run initial steps, is it so bad?

It might be good to revisit that option, indeed.

luke-hill:
Think the whole of this project needs a revisit and a version release

PR #410, opened 2019-04-10
Ensure working version of sqlite3 is used in scenarios

mvz:
## Summary

Force the use of sqlite3 version 1.3.x when running scenarios that test installing Rails and cucumber-rails. Since these scenarios break out of the bundle created by the gemspec, they might otherwise pick up later versions of sqlite3 by accident.

This is currently causing build failures with Rails 4.2 and 5.0, since they have no released version with support for sqlite3 1.4.

## Details

Add a gem line for sqlite3 to the relevant cucumber step.

## Motivation and Context

The Travis build is currently failing.

## How Has This Been Tested?

Running the full test suite for Rails 5.2. Travis will test the other versions.

## Types of changes

- Bug fix (non-breaking change which fixes an issue).

### DIFF
--- a/Appraisals
+++ b/Appraisals
@@ -1,17 +1,25 @@
 appraise "rails_4_2" do
   gem "railties", "~> 4.2.1"
+  gem "i18n", "< 1.6.0", platform: :mri_22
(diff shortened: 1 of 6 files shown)


mvz:
This now also fixes some other build issues.

PR #411, opened 2019-04-10
CI: Use 2.2.10, 2.4.6, 2.5.5, 2.6.2

olleolleolle:
## Summary

This PR updates the CI matrix to the latest patch releases.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -32,6 +32,10 @@ matrix:
       gemfile: gemfiles/rails_6_0.gemfile
     - rvm: 2.4.6
       gemfile: gemfiles/rails_6_0.gemfile
+    - rvm: 2.4.6
+      gemfile: gemfiles/rails_4_2.gemfile
+    - rvm: 2.5.5
+      gemfile: gemfiles/rails_4_2.gemfile
 
 before_install:
   - gem install --remote bundler -v '~> 1.17'
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,9 +1,9 @@
 rvm:
-  - 2.2.9
+  - 2.2.10
   - 2.3.8
-  - 2.4.5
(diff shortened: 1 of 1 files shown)


mvz:
I think this pulls in Bundler 2.0, opening a whole new can of worms.

.travis.yml
```diff
@@ -1,9 +1,9 @@
 rvm:
-  - 2.2.9
+  - 2.2.10
   - 2.3.8
-  - 2.4.5
-  - 2.5.3
-  - 2.6.0
+  - 2.4.6
```
luke-hill: for ref all of these don't need the teeny values. They get auto-set at runtime to be the latest

olleolleolle: Example where it does not: this was run yesterday `2.4` became `2.4.5`.
https://travis-ci.org/ruby/csv/jobs/520104049

luke-hill: I know occassionally if rvm isn't up to date it does that, but on that repo I couldn't find a `2.4`

olleolleolle: Doh! Thanks for reminding to clean my 👓 ! Apologies. I stand corrected.

You're right, and it often does not make any difference, and in the long run, `rvm` will catch up.

I'd be OK with changing the MRI versions to "no teeny values" in this PR.

_Personal note_: For me, `jruby` has been a core concern, and I want CI to run latest patch version of JRuby, to catch any issues with that release.

luke-hill: Just as a catch-up. I fully agree for any non-MRI that being fully explicit probably is the way to go. It was more of an observation than a non-blocking comment. But cheers for responding :)

olleolleolle: 👋 I would like to say thanks for your continued efforts in Cucumber. It's _awesome_!

PR #414, opened 2019-04-23
Additional Dependency / Cop Fixes

luke-hill:
## Summary

Additional rubocop auto-fixes (Most are done now), and began tackling the manual cops (But some are quite difficult / long-winded to fix)

## Details

As written above

## Motivation and Context

Makes the codebase better, and allows us to slowly remove the legacy / old code.
This also highlights a lot of bad-practices in our examples / tests that need addressing.

## How Has This Been Tested?

`bundle`
`rubocop`
`bundle exec rake`

## Types of changes

## Checklist:


luke-hill:
It would be good to discuss / raise Issues for items we want to change but need to discuss.

Few examples I found:
- Rewriting old tests, in what way?
- Rubocop rule overrides, which and why?
- Integrating rubocop performance / optimisation cops. When?

cucumber-rails.gemspec
```diff
@@ -3,31 +3,29 @@ $LOAD_PATH.unshift File.expand_path('lib', __dir__)
...
+  s.authors     = ['Aslak Hellesøy', 'Dennis Blöte', 'Rob Holland']
+  s.description = 'Cucumber Generator and Runtime for Rails'
   s.summary     = "#{s.name}-#{s.version}"
   s.email       = 'cukes@googlegroups.com'
-  s.homepage    = "http://cukes.info"
+  s.homepage    = 'http://cukes.info'
 
   s.license = 'MIT'
 
-  s.required_ruby_version = '>= 2.2'
```
olleolleolle: Good catch!

features/step_definitions/cucumber_rails_steps.rb
```diff
@@ -58,19 +56,26 @@ def gem(name, *args)
...
   end
 
   def fixture(path)
     File.expand_path(File.dirname(__FILE__) + "./../support/fixtures/#{path}")
   end
+
+  private
+
+  def rails6?
+    `bundle exec rails -v`.split("\n").last.strip.start_with?('Rails 6')
```
olleolleolle: Can one use  `.lines.last` here?

```suggestion
    `bundle exec rails -v`.lines.last.strip.start_with?('Rails 6')
```

features/support/legacy_web_steps_support.rb
```diff
@@ -283,7 +283,7 @@ def path_to(page_name)
 end
 
 World(NavigationHelpers)
-EOF
+    EOF
```
olleolleolle: This, doesn't it require the squiggly heredoc `<<~EOF` at its start?

lib/cucumber/rails.rb
```diff
@@ -1,4 +1,7 @@
-env_caller = File.dirname(caller.detect { |f| f =~ /\/env\.rb:/ }) if caller.detect { |f| f =~ /\/env\.rb:/ }
+if caller.detect { |f| f =~ /\/env\.rb:/ }
+  env_caller = File.dirname(caller.detect { |f| f =~ /\/env\.rb:/ })
+end
```
olleolleolle: Perhaps this can assign to `env_caller` the result of the `if` expression?

Why I say that: Now, the `env_caller` after line 3 is undefined, if we did not visit line 2.

features/step_definitions/cucumber_rails_steps.rb
luke-hill: Probably, I'll make the change and then go with it. For this manual one I just copied and pasted what was there beforehand.

features/support/legacy_web_steps_support.rb
luke-hill: So this entire file is 2 massive heredocs.

Heredoc 1 is: 3-250
Heredoc 2 is: 252-286

I've not tweaked anything here other than a tab, and it still all runs. However this entire file / set of steps should be just deleted because they're so bad. Aslak has put a massive comment at the top eluding to this.

lib/cucumber/rails.rb
luke-hill: It'll never be undefined, it'll be `nil` though, because of the way ruby lazy-evaluates local variables. I remember when I first learnt this it confused me.

```
a = 3
if a > 3
  b = 5
end

puts b => `nil` # This seems odd but it works.
```

.ruby-version
```diff
@@ -0,0 +1 @@
+ruby-2.2
```
luke-hill: @mvz I'm aware you said this caused an issue on your rbenv, but we really need to be running 2.2 locally at some point, so I'd argue this is worth keeping in, whilst we're supporting this low version.

On a side-note, are we planning on upping what versions we support?

mvz: > we really need to be running 2.2 locally at some point

Can you explain why? I think running 2.2 in CI is enough, while using whichever later version is convenient for local development.

luke-hill: Whilst ruby doesn't adhere strictly to semver, we should be developing things which work against 2.2 (Using the assumption that if they work on 2.2, they should work on 2.2+)

Extending that logic across the local/CI dev tree, then suggests we should be developing against 2.2. It is done in the cucumber-ruby repo, which is the closest mirror to this.

mvz: I don't see `.ruby-version` in the `cucumber-ruby` repository, and it is even explicitely ignored in their `.gitignore`: https://github.com/cucumber/cucumber-ruby/blob/master/.gitignore#L29. Maybe that makes sense to do here as well?

aslakhellesoy:
How do we test with different ruby versions when the file is there?

.ruby-version
luke-hill: Evidently this has changed recently. Mybad :(

If I add it to the gitignore is that ok?

luke-hill:
I just shift my ruby version. But I can see the preference is to ignore it so I'll do that. I was just arguing that the default should be the lowest version, but I'll cede here.

.ruby-version
mvz: No problem. Adding to gitignore is fine I think.

lib/cucumber/rails.rb
```diff
@@ -1,5 +1,8 @@
-env_caller = File.dirname(caller.detect { |f| f =~ /\/env\.rb:/ }) if caller.detect { |f| f =~ /\/env\.rb:/ }
-if env_caller
+called_from_env_rb = caller.detect { |f| f =~ /\/env\.rb:/ }
+
+if called_from_env_rb
+  env_caller = File.dirname(called_from_env_rb) if called_from_env_rb
```
mvz: I like this clean-up. There's a small nit: This final `if called_from_env_rb` isn't needed.

luke-hill:
Is this RtM @mvz - I did a couple of follow-up PR's (Much smaller), based off this I want to push up (Once this is in)

PR #418, opened 2019-04-25
Allow SQLite1.4+ in cucumber-rails

luke-hill:
## Summary

Allow SQLite1.4 in cucumber-rails, thus mitigating rails6 issues (Rails 6 has this as a dependency)

## Details

In the areas I could find I've amended the requirement to be more flexible

## Motivation and Context

It fixes rails6 builds failing on CI

## How Has This Been Tested?

Travis runs should hopefully pass (However I do admit I'm not sure I've fixed it in all the right areas.

## Types of changes

## Checklist:


luke-hill:
This fails on rails 5.0 and 4.2 (Due to a version mismatch, as expected for the SQLite gem)

luke-hill:
This ran and passed locally on all rails versions on Ruby 2.5. So I'm hoping this fixes part of the breakages!

README.md
```diff
@@ -75,8 +75,8 @@ Make sure you have a supported ruby installed, cd into your cucumber-rails repos
 
     gem install bundler
     bundle install
-    bin/install_geckodriver.sh
-    bin/install_webpacker.sh
+    ./bin/install_geckodriver.sh
+    ./bin/install_webpacker.sh
```
mvz: This `./` shouldn't be needed, right? Is there a circumstance where the old instructions are a problem?

README.md
```diff
@@ -86,25 +86,25 @@ With all dependencies installed, all specs and features should pass:
 
 ### Running Appraisal suite
 
-In order to test against multiple versions of key dependencies, the [Appraisal](https://github.com/thoughtbot/appraisal) is used to generate multiple gemfiles, stored in the `gemfiles/` directory. Normally these will only run on Travis; however, if you want to run the full test suite against all gemfiles, run the following commands:
+In order to test against multiple versions of key dependencies, the [Appraisal](https://github.com/thoughtbot/appraisal) gem is used to generate multiple gemfiles.
+
+These are stored in the `gemfiles/` directory. Normally these will only run on Travis; however, if you want to run the full test suite against all gemfiles, run the following command:
 
-    [bundle exec] appraisal install
```
mvz: I think this is needed to make sure all gems are actually installed at the start?

README.md
```diff
@@ -86,25 +86,25 @@ With all dependencies installed, all specs and features should pass:
...
     [bundle exec] appraisal rake test
 
 To run the suite against a named gemfile, use the following:
 
     [bundle exec] appraisal rails_4_2 rake test
 
 To remove and rebuild the different gemfiles (for example, to update a rails version or its dependencies), use the following:
+Note only a maintainer of the repo is likely to want to do this (As it will potentially change version dependencies).
```
mvz: :+1:

features/step_definitions/cucumber_rails_steps.rb
```diff
@@ -17,21 +17,24 @@ def install_cucumber_rails(*options)
       gem 'cucumber-rails', group: :test, require: false, path: File.expand_path('.').to_s
     end
 
-    gem 'sqlite3', '~> 1.3.13'
+    if rails6?
```
mvz: :+1:

README.md
luke-hill: It was for me, but I'm on ubuntu. It's my own fault really. I couldn't get yarn to update or rails6 to work, and then I manually fixed everything. When I re-checked this file I realised I hadn't ran the commands properly.

I can remove them if you'd prefer. Just helps me on Bionic run them effectively as they're shells.

README.md
luke-hill: When I was trying to fix things, this re-generates them. It's not like `bundle install` that doesn't regenerate if one is there. Which then causes rubocop / styling issues.

The gemfiles are currently under source control, so in theory you would only need to do the run command (Which would then bundle the specific gemfile).

I admit I'm a bit hazy on this, so I "could" be wrong. But I am fairly certain we shouldn't be advising anyone to run the install command as it changes all the sub-gemfiles

mvz: Hm .. if I don't run `appraisal install`, but just `appraisal rake test`, I get "Could not find i18n-0.9.5 in any of the sources", but if I do run `appraisal install`, I get an error because my lockfiles are out of date. :thinking: 

I'm going to read up on this a bit.

mvz: 1. I've asked a question about cleaning up the lockfiles: https://github.com/thoughtbot/appraisal/issues/153
2. I think we need to keep `appraisal install` because we need gems installed. Since we warn that this is for maintainers, the resulting changes to the gemfiles are not that much of a problem, I think.
3. We may need to exclude the gemfiles from rubocop since we'll be re-generating them at some point and they'll use double-quoted strings and such.
4. (Not really related to this PR) I think the `[bundle exec]` bits here can go because the whole point is that appraisal will activate different bundles while running.

luke-hill: There "ought" to be a command that uses the existing rules though and not re-installing them (like bundle does with bundle or bundle install).

I'll remove the files from rubocop tracking; what do you advise for the remainder. Should we wait to hear back from thoughtbot. As I'm not that knowledgable about Appraisals, I kind of got it working locally, but I didn't need to run the `install` command, but I may be doing something else unknowingly.

TL;DR - I don't really know what to do here.

luke-hill:
@olleolleolle you've done a load of work with this Appraisals stuff. I "think" I've got part of it working again now with rails6, but I don't really know the ins and outs. Can you maybe push some commits to this PR to get it all reading correctly, as I'm kinda half-guessing.

mvz:
@luke-hill How about rebasing just the 'technical' part of this, and postponing the documentation fixes to a separate pull request?

luke-hill:
Have done that now. Assuming this build passes again (Which it should), this should be RtM

PR #419, opened 2019-04-26
Fix up a few more rubocop offenses

luke-hill:
## Summary

Fixed up some more rubocop offenses

## Details

Fixed up LineLength cops (mainly), and a couple of other groups

I've also fixed some bugs in the rubocop config (As we were testing some of the `tmp` files, now we don't test anything that's generated into `tmp/` I've also added a new rule ignore for all spec block sizes (As most of them are likely to break the 25line limit)

## How Has This Been Tested?

`bundle exec rubocop`

## Types of changes

## Checklist:


.rubocop.yml
```diff
@@ -25,4 +25,5 @@ Style/RegexpLiteral:
 
 BlockLength:
   Exclude:
+    - './spec/**/*_spec.rb'
```
olleolleolle: :+1:

PR #420, opened 2019-05-03
Convert regex samples to better more up to date ones.

luke-hill:
## Summary

As written above, use better regex statements in sample features

## Motivation and Context

More tidying up of the test suite

## How Has This Been Tested?

`bundle exec cucumber`

## Types of changes

## Checklist:

This will slightly conflict with the other PR I made, but I'll rebase whichever one is approved last and then merge the 2nd one in.


xtrasimplicity:
Thanks, @luke-hill. That seems reasonable to me.

I'll review this further when I get a chance. :)

luke-hill:
Is this GTM?

luke-hill:
Re-ping @olleolleolle @xtrasimplicity

features/choose_javascript_database_strategy.feature
```diff
@@ -37,13 +37,13 @@ Feature: Choose javascript database strategy
       """
       Feature:
         Background:
-          Given I have created 2 widgets
+          When I create 2 widgets
```
xtrasimplicity: I think `Given` works better here, to be honest, as `when` usually denotes an action that you do _given_ a state rather than set the state itself.

features/choose_javascript_database_strategy.feature
```diff
@@ -56,16 +56,16 @@ Feature: Choose javascript database strategy
...
       """
-      Given /created? (\d) widgets/ do |num|
+      When(/^I create (\d+) widgets/) do |num|
         num.to_i.times { Widget.create! }
       end
 
-      Then /should have (\d) widgets/ do |num|
-        Widget.count.should == num.to_i
+      Then(/^I should have (\d+) widgets$/) do |num|
+        expect(Widget.count).to eq(num.to_i)
```
xtrasimplicity: This is good. I like the use of pronouns in steps - makes it more obvious that cucumber is about improving collaboration and makes it easier to adhere to business rules.

xtrasimplicity: The switch to `rspec-expectations` is good, too!

features/choose_javascript_database_strategy.feature
```diff
@@ -85,7 +85,7 @@ Feature: Choose javascript database strategy
       @javascript
       Feature:
         Background:
-          Given I have created 2 widgets
+          When I create 2 widgets
```
xtrasimplicity: Not so sure about this one, either. See above.

xtrasimplicity:
Sorry for the delay in getting to this, @luke-hill - things have been hectic at work. :)

features/choose_javascript_database_strategy.feature
luke-hill: if you look at what is happening, a `create` action is being invoked, which suggests it is an action. That was my 2 cents.

xtrasimplicity: Yep, I understand that. Just not sure about the lack of `Given`. I've always understood that `given` sets the initial state and `when` manipulates the state in a way that is verified by subsequent `then`steps.

I may have interpreted wrongly, though. :)

luke-hill:
Merging this in (Have tested that the branch earlier passed CI when adding a phantom file)

PR #423, opened 2019-05-22
Remove Aruba reporting monkey-patch

mvz:
## Summary

Remove Aruba reporting monkey-patch

## Details

Removes code from Cucumber-Rails that tests for ARUBA_REPORT_DIR and overrides Aruba's reporting code. Also removes `prepare_aruba_report` calls from the step definitions.

## Motivation and Context

Aruba reporting is deprecated and broken with Cucumber 2 and up, and Cucumber-Rails requires Cucumber 3. Therefore, it is safe to say that no-one is using this feature for running Cucumber-Rails' test suite.

## How Has This Been Tested?

I ran `ARUBA_REPORT_DIR=tmp bundle exec rake` without this change and it fails, even after adding `bcat` as a dependency, with the message:

```
undefined method `line' for #<Cucumber::Core::Test::Case: features/allow_rescue.feature:20> (NoMethodError)
```

## Types of changes

Basically a cleanup. Removes something that was broken.

## Checklist:

### DIFF
--- a/features/step_definitions/cucumber_rails_steps.rb
+++ b/features/step_definitions/cucumber_rails_steps.rb
@@ -2,14 +2,12 @@ Given('I have created a new Rails app and installed cucumber-rails, accidentally
   rails_new
   install_cucumber_rails :not_in_test_group
   create_web_steps
-  prepare_aruba_report
 end
 
 Given('I have created a new Rails app and installed cucumber-rails') do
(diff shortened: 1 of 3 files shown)


mvz:
> Do we not have any tests for this?

No, it was just there for cucumber-rails devs to use I guess.

PR #424, opened 2019-05-22
Drop Ruby 2.2 support

mvz:
## Summary

Drop Ruby 2.2 support, target 2.3 and up.

## Details

* Target 2.3+ in gemspec, RuboCop config
* Ensure sqlite3 is available for running just `bundle exec rake`
* Fix new RuboCop offense

## Motivation and Context

Ruby 2.2 is ancient and was removed from the build set. Cucumber-rails should not claim support for it if that support isn't tested.

## How Has This Been Tested?

CI.

## Types of changes

- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected).

## Checklist:


cucumber-rails.gemspec
```diff
@@ -27,13 +29,14 @@ Gem::Specification.new do |s|
   s.add_development_dependency('rspec', '~> 3.6')
   s.add_development_dependency('rubocop', '~> 0.66.0')
   s.add_development_dependency('rubocop-rspec', '~> 1.32.0')
+  s.add_development_dependency('sqlite3', '~> 1.3')
```
luke-hill: This "may" cause issues with appraisals, but I know without this you can't just run a standard thing on your own.

It's a bit of chicken and egg scenario I admit. Hopefully nothing bad happens.

Appraisals
```diff
@@ -1,3 +1,5 @@
+# frozen_string_literal: true
+
```
deivid-rodriguez: You can now remove all the `mri_22` requirements below.

mvz: @deivid-rodriguez Done!

olleolleolle:
This is so great, a great win!

PR #413, opened 2019-04-18
Updated Dependencies for Cucumber Rails (For 2019+)

luke-hill:
## Summary

A generic bump of most things (Gems / Styles e.t.c.)

## Details

- Development dependencies have all gone up 2+ years; some were archaic. We should be definitely developing against stuff from 2016+!
- A first pass of rubocop auto-fixes has been done. Just tackling the low-hanging fruit. About 50% has been auto-fixed. There is still a lot of obvious/easy stuff to fix, but I didn't want to bloat the PR.
- Travis tweaks: Ruby 2.2 has been removed and some exclusions have been added so we're not testing the combinations that we shouldn't be (Happy to alter this if reviewers wish me to)

## Motivation and Context

This brings Cucumber-Rails into the 2019's

## How Has This Been Tested?

`bundle exec rake`

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue).
- [ ] New feature (non-breaking change which adds functionality).
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected).

## Checklist:

- [ ] I've added tests for my code.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.

PR #434, opened 2019-07-31
Feature/doc cop review

luke-hill:
## Summary

Fix up docs / cops

## Details

Final of 3 small tix to fix up docs / cops (Partially)

## Motivation and Context

Clean up codebase

## How Has This Been Tested?

CI

## Types of changes

## Checklist:


.rubocop.yml
```diff
@@ -20,6 +20,7 @@ Style/Documentation:
   Enabled: false
 
 Metrics/LineLength:
+  Max: 200
```
mvz: This will make RuboCop force lots of conditionals to use postfix syntax, so better keep this shorter, say at 92 or 100.

CHANGELOG.md
```diff
@@ -6,6 +6,38 @@ Please see [CONTRIBUTING.md](https://github.com/cucumber/cucumber/blob/master/CO
...
+   [luke-hill](https://github.com/luke-hill))
+   
+### Fixed
+
+ * Various rubocop / styling issues that had built up over time
+  ([#413](https://github.com/cucumber/cucumber-rails/pull/413) /
+  [#414](https://github.com/cucumber/cucumber-rails/pull/414) /
+  [#416](https://github.com/cucumber/cucumber-rails/pull/416) /
+  [#419](https://github.com/cucumber/cucumber-rails/pull/419) /
+  [#419](https://github.com/cucumber/cucumber-rails/pull/420) /
```
mvz: The `[#419]` should be `[#420]` in this line.

CHANGELOG.md
```diff
@@ -6,6 +6,38 @@ Please see [CONTRIBUTING.md](https://github.com/cucumber/cucumber/blob/master/CO
...
+ * Begin to update the core runtime / development dependencies to something a little more recent
+   ([#431](https://github.com/cucumber/cucumber-rails/pull/431) /
+   [#432](https://github.com/cucumber/cucumber-rails/pull/432) /
+   [mvz](https://github.com/mvz) / 
+   [luke-hill](https://github.com/luke-hill))
+   
+### Fixed
+
+ * Various rubocop / styling issues that had built up over time
+  ([#413](https://github.com/cucumber/cucumber-rails/pull/413) /
```
mvz: Pull request #413 is about dependencies so doesn't belong here.

dev_tasks/yard.rake
```diff
@@ -8,9 +8,7 @@ API_DIR = File.join(SITE_DIR, 'api', 'cucumber-rails', 'ruby', 'yardoc')
 
 namespace :api do
   file :dir do
-    unless File.directory?(SITE_DIR)
-      raise "You need to git clone git@github.com:cucumber/cucumber.github.com.git #{SITE_DIR}"
-    end
+    raise "You need to git clone git@github.com:cucumber/cucumber.github.com.git #{SITE_DIR}" unless File.directory?(SITE_DIR)
```
mvz: This is where RuboCop does the wrong thing due to line length.

(Once upon a time there was a separate setting for this particular cop, independent of line length. The more I think about it, the more I'm convinced this setting should return.)

.rubocop.yml
marnen: Why do we need this at all? The syntax of the Ruby language doesn’t generally lead to long lines anyway, so I tend to think a style cop for it would be excessive (and difficult for those few but nonzero cases where the long line is the preferable option).

luke-hill: See SHA - Standardisation across repos https://github.com/cucumber/cucumber-rails/pull/434/commits/40fc5b25479c00ccdd55449663e90b529192dafd

CHANGELOG.md
luke-hill: It was mainly just doing rubocop auto-fixes. It also fixed up some dependencies yes. I'll add it as a separate entry if you'd prefer

.rubocop.yml
mvz: Well, I think the reasoning in `cucumber-ruby` is wrong. Line length in this github issue view is a lot less than 200.

I think @marnen is asking why this cop is not simply disabled? I'm afraid there are still people who needlessly put a lot of stuff on a single line. This is bad for readability so some check is called for.

CHANGELOG.md
mvz: I'll check this again.

mvz: I think it makes most sense to just link to this pull request with both changelog entries.

.rubocop.yml
marnen: @mvz Personally I’d rather catch that in code review. There can be 30-character lines that are too long and (occasionally) 120-character lines that are not too long. I don’t think a hard character limit does any good.

mvz: There could always be exceptions, but for the majority of cases a hard character limit is useful for automating the boring part of code review.

I would never flag a 30-character line as just 'too long', but rather something else would be wrong with it, like 'too complex', or 'multiple statements on one line'.

marnen: @mvz If there can always be exceptions, then is this suitable for cop automation? I prefer to automate only those rules that we absolutely never want to violate, because I don’t like merging red code into the project.

mvz: Exceptions will have to be marked as such to avoid the build failing.

mvz: Note that I think exceptions for line length are highly unlikely, and I think the reasons for making such an exception would have to be really strong indeed.

marnen: @mvz As do I. But if the cop is set to 200 characters, we might as well not have it at all: no one would ever submit a pull request with a line that long in the first place, and if they did, it would be immediately visually obvious.

dev_tasks/yard.rake
marnen: @mvz Thanks, this is a great example of a 122-character line that isn't too long. I wouldn't want to break up the string literal just for the sake of breaking it up.

.rubocop.yml
mvz: @marnen I'm not arguing for 200 chars.

dev_tasks/yard.rake
mvz: @marnen the string literal isn't the reason this line is so long. In the original, the longest line (12) was 95 characters. In any case, I'd still consider splitting it up since it has actual code at the end (`SITE_DIR`).

.rubocop.yml
mvz: @luke-hill if you prefer to keep it at 200 for consistency, I'm OK with that.

luke-hill: If the cucumber-ruby or other ruby-ish ones change it. I'll change it here. I think that's a debate for another day. I'm for standardisation

mvz: @luke-hill agreed.

dev_tasks/yard.rake
marnen: ...whereas for me, doing so would be a perfect example of pointless use of character count. :) I think it’s fine as it stands.

PR #436, opened 2019-08-06
Cut v1.8

luke-hill:
Release v1.8

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,16 @@
 Please see [CONTRIBUTING.md](https://github.com/cucumber/cucumber/blob/master/CONTRIBUTING.md) on how to contribute to Cucumber.
 
-## [master](https://github.com/cucumber/cucumber-rails/compare/v1.7.0...master) (Not yet released)
+## [master](https://github.com/cucumber/cucumber-rails/compare/v1.8.0...master) (Not yet released)
+
+### Changed
+
+ * 
+   
+### Fixed
+
+ * 
+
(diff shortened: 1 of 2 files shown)


olleolleolle:
@luke-hill Can you release the gem to RubyGems too?

luke-hill:
I think @xtrasimplicity needs to do this. I don't have push capabilities. I'll message some people on slack about this though

xtrasimplicity:
I'll do that in a few mins. @luke-hill

xtrasimplicity:
@luke-hill Done. It's about 1/10th of the size of the last release - is that to be expected?
https://rubygems.org/gems/cucumber-rails/versions/1.8.0

luke-hill:
Unsure.

It could be a rubygems packaging thing perhaps?

https://github.com/cucumber/cucumber-rails/compare/v1.7.0..v1.8.0 the diff shows a net reduction in code, but most of that isn't library code it's yml from rubocop

I do know changing some of the verbose globals will save a "bit" of space. But I'm not 100% sure.

Maybe try rebuilding the 1.7 gem and check the size again? It does sound a bit worrying. Could also check another instance of the 1.8 gem?

luke-hill:
EDIT: False alarm. Nothing to worry about. I've removed 9/10ths of the crud.

We had a couple of physical gems that were bundled as part of the cucumber-rails gem. Now they no longer are!

xtrasimplicity:
Awesome! Great work! :)

PR #352, opened 2018-02-01
Remove Rack::Test::Methods from the default World.

MonsieurDart:
For some reason, the Cucumber `World` instantiated by default by Cucumber-Rails includes the module `Rack::Test::Methods`. This will have the effect of overriding all the default Rails IntegrationTest helpers (`get`, `post`, etc.). Is there a reason for that?

Could we imagine removing this `include` from the default configuration? This seems a right direction, IMHO, for two reasons: 

- It is very easy to add this from the user's Cucumber `env.rb` file, with: `World(Rack::Test::Methods)`.
- It does not seem possible to force the usage of the Rails helpers, instead of the Rack ones (but I can be wrong, I'm quite of a Rails newbie).

Any opinions on that?

PR #440, opened 2019-08-26
Feature/configure Rack::Test methods

luke-hill:
## Summary

Allow `Rack::Test Methods` to be either included or not included (mixed in), to the cucumber world.

## Details

Add an environment variable `CR_REMOVE_RACK_TEST_HELPERS=true` configuration switch to include or exclude these methods.

So far I have not documented these as I'm unsure where is best.

## Motivation and Context

Supersedes #352 (Old PR)

## How Has This Been Tested?

CI and a new feature test for all conditionals

## Types of changes

## Checklist:


lib/cucumber/rails/world.rb
```diff
@@ -7,10 +7,20 @@
   # Test Unit not found
 end
 
+module Cucumber
+  module Rails
+    class << self
+      def remove_rack_test_helpers
+        ENV['CR_REMOVE_RACK_TEST_HELPERS'] == "true"
```
olleolleolle: Perhaps you can use https://api.rubyonrails.org/classes/ActiveModel/Type/Boolean.html this coercion thing, since this is a Rails add-on, we have Rails available.

luke-hill: I usually dislike these things, but as this is a free gain. I'll fix this up (If we didn't have rails I'd just do a `casecmp == 0` or something

olleolleolle:
(The CI tests currently fail.)

luke-hill:
@olleolleolle are you happy for this to be merged?

olleolleolle:
@luke-hill I really like the functionality, but I am torn about the name - the configuration logic we are amening is about "include this, but only if we should"

Perhaps the name of the supporting method could be changed, keeping the ENV variable setting's name?

luke-hill:
That's a valid point. And we "could" do that. However I'd rather do that in 2 phases. Because we currently always include it. So for us to remove it completely we would need to do an interim release beforehand warning users. Which I hadn't planned on doing.

If we go from always including it to only including it if the user does something we'd need to give them plenty of notice. This is the way that gives the users a little flexibility.

olleolleolle:
My phrasing was wrong: I meant to say that the current _way_ of doing the skipping etc is the right one.

When reading the implementation, though, the `unless ...` could be turned into a more-readable `if`.

- turn the logic around in the body of the supporting method (now named `remove_rack_test_helpers`
- proposed rename `.remove_rack_test_helpers` => `.include_rack_test_helpers?`

PR #447, opened 2019-11-04
Add project metadata to the gemspec

orien:
Add `bug_tracker_uri`, `changelog_uri`, `documentation_uri`, `mailing_list_uri`, and `source_code_uri` to the gemspec metadata.

These [project metadata](https://guides.rubygems.org/specification-reference/#metadata) will facilitate easy access to project information. The URI will be available on the [Rubygems project page](https://rubygems.org/gems/cucumber-rails), via the Rubygems API, and the `gem` and `bundle` command-line tools with the next release.

Also update the homepage URL.

### DIFF
--- a/cucumber-rails.gemspec
+++ b/cucumber-rails.gemspec
@@ -18,8 +18,7 @@ Gem::Specification.new do |s|
     'changelog_uri' => "https://github.com/cucumber/cucumber-rails/blob/v#{s.version}/CHANGELOG.md",
     'documentation_uri' => 'https://cucumber.io/docs',
     'mailing_list_uri' => 'https://groups.google.com/forum/#!forum/cukes',
-    'source_code_uri' => "https://github.com/cucumber/cucumber-rails/tree/v#{s.version}",
-    'wiki_uri' => 'https://github.com/cucumber/cucumber-rails/wiki'
+    'source_code_uri' => "https://github.com/cucumber/cucumber-rails/tree/v#{s.version}"
   }
 
   s.add_runtime_dependency('capybara', ['>= 2.12', '< 4'])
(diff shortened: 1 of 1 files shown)


cucumber-rails.gemspec
```diff
@@ -9,10 +9,19 @@ Gem::Specification.new do |s|
...
 
   s.license = 'MIT'
 
+  s.metadata = {
+    'bug_tracker_uri' => 'https://github.com/cucumber/cucumber-rails/issues',
+    'changelog_uri' => "https://github.com/cucumber/cucumber-rails/blob/v#{s.version}/CHANGELOG.md",
+    'documentation_uri' => 'https://cucumber.io/docs',
+    'mailing_list_uri' => 'https://groups.google.com/forum/#!forum/cukes',
+    'source_code_uri' => "https://github.com/cucumber/cucumber-rails/tree/v#{s.version}",
+    'wiki_uri' => 'https://github.com/cucumber/cucumber-rails/wiki'
```
olleolleolle: This one should perhaps be omitted - we are in the process of moving hard-to-keep-current wiki content to the docs site.

orien: I notice there's an (old and non-working) value recorded on [Rubygems](https://rubygems.org/gems/cucumber-rails):

http://wiki.github.com/cucumber/cucumber-rails

I wonder if setting this to `nil` would remove it.

```suggestion
    'wiki_uri' => nil
```

orien: I notice there's an (old and non-working) value recorded on [Rubygems](https://rubygems.org/gems/cucumber-rails):

http://wiki.github.com/cucumber/cucumber-rails

I wonder if setting this to `nil` would remove it.

```suggestion
    'wiki_uri' => nil
```

olleolleolle: The absence of the `'wiki_uri` key would be best.

orien: Removed in 0b036ea

PR #448, opened 2019-11-04
Do not require RSpec if not already loaded

### LINKED ISSUE #446: Check for RSpec::Matchers rather than requiring it in begin/rescue block
## Summary

I am working on a project where MiniTest and `rspec-mocks`, `rspec-expectations` are both available in the Gemfile. However we have RSpec disabled as we do not use RSpec assertions in Cucumber. This is causing an issue with `cucumber-rails` where it tries to load RSpec expectations and matchers regardless of them being necessary.

## Expected Behavior

Tests run fine without RSpec getting loaded when the Gemfile has `require: false` for those libraries.

## Current Behavior

`assert_equal` with capybara fails due to `rspec-expectations` changing its behavior from vanilla `minitest`.

## Possible Solution

Changing line the [lib/rspec.rb](https://github.com/cucumber/cucumber-rails/blob/master/lib/cucumber/rails/rspec.rb#L5-L22) file to check for `defined?(RSpec::Mock)` would allow for these two libraries to co-exist without causing a conflict.

jsmestad:
## Summary

Fixes #446 

## Types of changes

## Checklist:

### DIFF
--- a/lib/cucumber/rails/rspec.rb
+++ b/lib/cucumber/rails/rspec.rb
@@ -1,22 +1,18 @@
 # frozen_string_literal: true
 
+
 require 'cucumber/rails/world'
 
-begin
-  require 'rspec/rails/matchers'
-
+if defined?(RSpec::Matchers)
   [Cucumber::Rails::World, ActionDispatch::Integration::Session].each do |klass|
     klass.class_eval do
       include RSpec::Matchers
     end
   end
-rescue LoadError
-  require 'spec/expectations'
-  require 'spec/rails'
-
+else
   [Cucumber::Rails::World, ActionDispatch::Integration::Session].each do |klass|
     klass.class_eval do
-      include Spec::Matchers
+      include Spec::Matchers if defined?(Spec::Matchers)
       include Spec::Rails::Matchers if defined?(Spec::Rails::Matchers)
     end
   end


luke-hill:
Actually having re-reviewed the codebase. If you want to add a 1liner to include the extra file. I'm happy with that as a bug request fix. But altering the requires would be a feature change (which is fine), but I'd want that tweaking in other files ideally.

Also there's some argument it might be a major change (Again not a problem), but would need treating accordingly.

jsmestad:
> Whilst the file already has conditionals. Given these are procedural loaders I'd rather remove conditionals.
> 
> Maybe look into making another file for minitest?

We could do something to the inverse maybe to avoid the breaking change? Maybe something like: if minitest is loaded and RSpec is not, then dont require RSpec?

olleolleolle:
@luke-hill Would that translate to: change the `else` to another `if defined?(...)` conditional around the second `[Cucumber::Rails::World, ActionDispatch::Integration::Session].each do |klass|` ?

luke-hill:
Possibly. But I think a better approach would be to have a minitest initializer and an rspec one. Because at the moment, the rspec one (named as such), is controlling minitest because we (perhaps incorrectly), assumed that if one existed, the other didn't. And one always existed.

In the truth table, there's 4 situations. We need to cover each of them ideally. The OP did suggest we could cover 2 of the use cases (1 is loaded and the other isn't), by just checking for each of them

luke-hill:
@jsmestad any thoughts on this?

jsmestad:
@luke-hill I think the request became larger than what I was expecting and had to move onto other tasks. Since then we just switched to RSpec completely to avoid the issue altogether.

I agree with the approach, but the time involved was going to be too much for me to take on during year-end scramble.

luke-hill:
I'll kill the PR @jsmestad but leave the issue open so someone else can come into it. I'm thinking this will be hard to fix in a regular PR, but we could make a brute force change and cut a 3.0 release and just say

"cucumber rails always had an assumption you would use rspec, but now you need to explicitly tell it you want to use rspec, or minitest"

PR #453, opened 2020-01-07
Feature/enable cucumber 4

luke-hill:
## Summary

Draft attempt at making Cucumber 4 work on cucumber rails

## How Has This Been Tested?

CI

## Types of changes

## Checklist:

### DIFF
--- a/cucumber-rails.gemspec
+++ b/cucumber-rails.gemspec
@@ -22,7 +22,7 @@ Gem::Specification.new do |s|
   }
 
   s.add_runtime_dependency('capybara', ['>= 2.12', '< 4'])
-  s.add_runtime_dependency('cucumber', ['>= 3.0.2', '< 4'])
+  s.add_runtime_dependency('cucumber', ['>= 3.0.2', '< 5'])
   s.add_runtime_dependency('mime-types', ['>= 2.0', '< 4'])
   s.add_runtime_dependency('nokogiri', '~> 1.8')
   s.add_runtime_dependency('rails', ['>= 4.2', '< 7'])


luke-hill:
cc/ @stripedpumpkin

luke-hill:
Currently cucumber 4-rc3 uses invalid version of gherkin (8.2.1), need to wait until gherkin 9 / 10 are released inside cucumber4 release candidates before re-attempting to fix this branch

cucumber-rails.gemspec
```diff
@@ -22,10 +22,10 @@ Gem::Specification.new do |s|
   }
 
   s.add_runtime_dependency('capybara', ['>= 2.12', '< 4'])
-  s.add_runtime_dependency('cucumber', ['>= 3.0.2', '< 4'])
+  s.add_runtime_dependency('cucumber', ['4.0.0.rc3'])
```
vugryumov-mdsol: @luke-hill AFAIK the version is `4.0.0.rc.3`. And might as well bump it to rc.4 while you at it. 👍

deivid-rodriguez:
I'm going to rebase this PR and see how it's doing.

deivid-rodriguez:
This seems to work just fine. As a matter of fact, cucumber 4 changelog claims "no breaking changes", and that seems right :)

Questions now about this PR:

* Given the stability of this major release, do we want to change the upper bound to `< 5` to allow minor 4.x releases without having to release counterparts in `cucumber-rails`?

* Was dropping cucumber 3.0.2 support and bumping the lower bound to cucumber 3.1 intentional?

mathieujobin:
unless there is a good reason to drop support for cucumber 3.0, I would not change the lower bound

good news on cucumber 4.0 not having any breaking changes.

deivid-rodriguez:
Agreed, I'll remove that bit from the PR. What about the upper bound? Does `< 5` make sense to you?

mvz:
> Does < 5 make sense to you?

I'm not that 'you', but it makes sense to me.

deivid-rodriguez:
Haha, great @mvz. I updated the PR. If it's ok with you, I'll merge it after CI.

luke-hill:
Probably should have been some doc updates for this as well, as the cucumber load order has been changed in cucumber4.

deivid-rodriguez:
Is that something that should be documented here, or in `cucumber-ruby`? How does it affect `cucumber-rails`?

By they way, the link to your blog post in https://github.com/cucumber/cucumber-ruby/blob/ac8d89bc739d39200f82b6bed1ad9fe7273f81ac/CHANGELOG.md#400rc2 seems broken.

luke-hill:
Yep, it was never confirmed. I spoke to Vincent about it this morning, but no idea what's going to happen.

I think having a small item here as well would help, because rails has its own autoloader, so it might just be worth a small footnote. But it's not mandatory.

PR #457, opened 2020-01-10
Switch to prepend instead of alias chaining methods

### LINKED ISSUE #454: Use `prepend` instead of alias method chaining
In multiple places alias method chaining is used instead of prepend. Since cucumber-rails doesn't work with version of Ruby prior to 2.0, these should be updated to the more semantically correct prepend.

wagenet:
Fixes #454


wagenet:
Looks like the test failures are unrelated.

wagenet:
@luke-hill I can definitely change it. What sort of naming would you prefer?

luke-hill:
your namespacing has an Extensions module. So either add a new folder for that or remove that module.

wagenet:
@luke-hill what about the other stuff in `lib/cucumber/rails/action_controller.rb`? Previously, it had things from two different modules.

luke-hill:
Things aren't great here. So if you want to fix up more stuff, go nuts. I'm just here to make sure we make stuff better ideally with each PR.

luke-hill:
ping @wagenet

wagenet:
@luke-hill sorry for the delay. I think it is ready now. Let me know if you need anything else.

wagenet:
@luke-hill I don't think the CI failures are my fault, but please let me know if I'm mistaken.

lib/cucumber/rails/action_dispatch.rb
```diff
@@ -0,0 +1,21 @@
...
+end
+
+module Cucumber
+  module Rails
+    module ActionDispatch
+      module ShowExceptions
+        def call(env)
+          env['action_dispatch.show_detailed_exceptions'] = !ActionController::Base.allow_rescue
+          env['action_dispatch.show_exceptions'] = !env['action_dispatch.show_detailed_exceptions']
+          super
```
luke-hill: this should be `super(env)` From previous logic

wagenet: Calling `super` without arguments passes all of the original arguments. I'm pretty sure that it will pass the modified `env` object, but given the potential for confusion here, I'll update it.

wagenet:
@luke-hill I'm seeing lots of stuff around deprecations in the failures.

lib/cucumber/rails/action_dispatch.rb
luke-hill: It will normally, but because we're doing a large refactor, lets be consistent.

luke-hill:
The failures aren't anything to worry about if they're rails6. They changed something between 6.0.1 and 6.0.2 that we've yet to hear back / diagnose

wagenet:
@luke-hill ok, looks like we're probably ok now.

aslakhellesoy:
Hi @wagenet,

Aslak Hellesøy
Creator of Cucumber

PR #462, opened 2020-03-26
Skip cache_classes warning when using Spring

janko:
## Summary

Avoid printing out warning when `cache_classes` needs to be `false`.

## Details

The `cache_classes` setting needs to be `false` when Spring is used. In this
case cucumber-rails doesn't have to print out the warning.

## Motivation and Context

See https://github.com/rails/spring/issues/598#issuecomment-528121218.

## How Has This Been Tested?

I'm not sure whether this change requires tests, it probably doesn't make sense
to pull in Spring just for that.

## Types of changes

## Checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,9 +12,10 @@ Please see [CONTRIBUTING.md](https://github.com/cucumber/cucumber/blob/master/CO
 
 ### Fixed
 
- * 
+ * Avoid printing a warning about `config.cache_classes` being set to `false` when
+   Spring is used ([#462](https://github.com/cucumber/cucumber-rails/pull/462) [janko])
 
-## [v2.0.0](https://github.com/cucumber/cucumber-rails/compare/v1.8.0...v2.0.0) (2019-09-013)
+## [v2.0.0](https://github.com/cucumber/cucumber-rails/compare/v1.8.0...v2.0.0) (2019-09-13)
 
 ### New Features
 
--- a/lib/cucumber/rails.rb
+++ b/lib/cucumber/rails.rb
@@ -15,7 +15,7 @@ if called_from_env_rb
 
   unless Rails.application.config.cache_classes || defined?(Spring)
     warn "WARNING: You have set Rails' config.cache_classes to false
-    or are using Spring. This is known to cause problems
+    (Spring needs cache_classes set to false). This is known to cause problems
     with database transactions. Set config.cache_classes to true if you want to use transactions."
   end
 
--- a/lib/cucumber/rails.rb
+++ b/lib/cucumber/rails.rb
@@ -13,7 +13,7 @@ if called_from_env_rb
   require 'cucumber/rails/action_dispatch'
   require 'rails/test_help'
 
-  unless Rails.application.config.cache_classes
+  unless Rails.application.config.cache_classes || defined?(Spring)
     warn "WARNING: You have set Rails' config.cache_classes to false
     (most likely in config/environments/cucumber.rb). This setting is known to cause problems
     with database transactions. Set config.cache_classes to true if you want to use transactions."
--- a/lib/cucumber/rails.rb
+++ b/lib/cucumber/rails.rb
@@ -15,7 +15,7 @@ if called_from_env_rb
 
   unless Rails.application.config.cache_classes || defined?(Spring)
     warn "WARNING: You have set Rails' config.cache_classes to false
-    (most likely in config/environments/cucumber.rb). This setting is known to cause problems
+    or are using Spring. This is known to cause problems
     with database transactions. Set config.cache_classes to true if you want to use transactions."
   end
 


lib/cucumber/rails.rb
```diff
@@ -13,7 +13,7 @@
   require 'cucumber/rails/action_dispatch'
   require 'rails/test_help'
 
-  unless Rails.application.config.cache_classes
+  unless Rails.application.config.cache_classes || defined?(Spring)
     warn "WARNING: You have set Rails' config.cache_classes to false
```
luke-hill: ... or are using Spring

olleolleolle: Ah, that becomes a more helpful message to the user, by including a note on Spring.

lib/cucumber/rails.rb
```diff
@@ -13,7 +13,7 @@
...
     (most likely in config/environments/cucumber.rb). This setting is known to cause problems
     with database transactions. Set config.cache_classes to true if you want to use transactions."
```
olleolleolle:
```suggestion
  unless Rails.application.config.cache_classes || defined?(Spring)
    warn "WARNING: You have set Rails' config.cache_classes to false
    or are using Spring. This setting is known to cause problems
    with database transactions. Set config.cache_classes to true if you want to use transactions."
```

Did you mean like this, @luke-hill ?

luke-hill:
@janko could you write a small changelog update for this then I'll merge it.

lib/cucumber/rails.rb
janko: I'm not sure what exactly you mean here, because the `cache_classes` settings is still explicitly set to `false` when Spring is used (Spring doesn't change this setting, it only implicitly requires it to be set to `false`).

janko: Oh, I see now that you've updated this. Well, as I said, I'm not sure "you have set `cache_classes` to false *or* you are using Spring" is accurate, because when you're using Spring the `cache_classes` setting still needs to be explicitly set to `false`. You *can* set it to `true`, and Spring won't complain about it, but then the code won't reload properly.

janko:
@luke-hill Added the changelog entry 👍

lib/cucumber/rails.rb
olleolleolle: Alternative formulations, to be accurate, yet still keep the suggestion that Spring could be the reason?

- (perhaps by using Spring)
- (running Spring needs cache_classes set to false)

luke-hill: That's fine. Whatever you think is best I'm happy with.

CHANGELOG.md
```diff
@@ -12,9 +12,10 @@ Please see [CONTRIBUTING.md](https://github.com/cucumber/cucumber/blob/master/CO
 
 ### Fixed
 
- * 
+ * Avoid printing a warning about `config.cache_classes` being set to `false` when
+   Spring is used ([#462](https://github.com/cucumber/cucumber-rails/pull/462) [janko])
 
-## [v2.0.0](https://github.com/cucumber/cucumber-rails/compare/v1.8.0...v2.0.0) (2019-09-013)
+## [v2.0.0](https://github.com/cucumber/cucumber-rails/compare/v1.8.0...v2.0.0) (2019-09-13)
```
olleolleolle: Thanks, :sparkles: good fix!

lib/cucumber/rails.rb
```diff
@@ -13,9 +13,9 @@
   require 'cucumber/rails/action_dispatch'
   require 'rails/test_help'
 
-  unless Rails.application.config.cache_classes
+  unless Rails.application.config.cache_classes || defined?(Spring)
     warn "WARNING: You have set Rails' config.cache_classes to false
-    (most likely in config/environments/cucumber.rb). This setting is known to cause problems
+    or are using Spring. This is known to cause problems
```
olleolleolle:
```suggestion
    (Spring needs cache_classes set to false). This is known to cause problems
```

janko:
Anything else I can do to help this get merged?

aslakhellesoy:
Hi @janko,

Aslak Hellesøy
Creator of Cucumber

luke-hill:
@janko sorry mybad. But olle helped you out. I just forgot.

olleolleolle:
And, thanks for taking the time to improve this!

lib/cucumber/rails.rb
Hirurg103: Running cucumber tests with spring and `config.cache_classes = true` in rails 6 fails with `reloading is disabled` error:
```
~/.rbenv/versions/2.6.5/lib/ruby/gems/2.6.0/gems/activesupport-6.0.3.2/lib/active_support/dependencies/zeitwerk_integration.rb:14:in `rescue in block in clear':
reloading is disabled because config.cache_classes is true (RuntimeError)
```

The following warning message looks more clear for me (what do you think, guys?):
```text
WARNING: You have set Rails' config.cache_classes to false.
This is known to cause problems with database transactions.
Set config.cache_classes to true or run cucumber with Spring if you want to use transactions.
See https://github.com/jonleighton/spring-commands-cucumber on how to configure Spring for cucumber.
```

BTW the [original warning message](https://github.com/cucumber/cucumber-rails/blob/e897d6e012d85728cb3f06209dd5a5f3e2a01d01/lib/cucumber/rails3.rb#L15) was telling about issues with **using transactional fixtures** when `cache_classes` is set to `false`. And provides a link to the original issue (https://rspec.lighthouseapp.com/projects/16211/tickets/165) in a private project. In 5c7030a1 the warning message has been rephrased that disabling cache for classes causes problems with **database transactions**. Does it mean DatabaseCleaner's `:transaction` strategy under "database transactions"? Or transactional fixtures? Does this problem still persists? It would be good to copy the original issue from the private project to [cucumber-rails GH issues](https://github.com/cucumber/cucumber-rails/issues) and add link to it to the warning. What do you think @aslakhellesoy ?

luke-hill: @Hirurg103 - When this PR was being formulated rails6 testing (On this repo), was still in its infancy, as such it's likely that this behaviour may have been modified by rails and our assumptions previously were correct but now may not be.

Also this error is coming from AS, a rails gem, which we don't control. We almost never throw an error for configuration in cucumber-rails, so you're probably best opening an issue on the rails/activesupport area.

PR #464, opened 2020-05-30
Fix CI

### LINKED ISSUE #461: CI Failures (Rails 6.0.1)
## Summary

Fix the issues on Rails 6 on CI

## Expected Behavior

CI passes

## Current Behavior

CI fails
## Possible Solution

Created an issue in Rails (Unsure if it will gain traction) - https://github.com/rails/rails/issues/38502

## Steps to Reproduce (for bugs)

1. Check CI with rails 6.0.0 - It passes
2. Check CI with rails 6.0.1 - It fails
3.
4.

## Context & Motivation

Get CI green again

## Your Environment

CI here

deivid-rodriguez:
## Summary

Just trying to get the CI green. This PR closes #461.

## Details

CI is broken on Rails 6. This PR introduces a number of changes to get CI green again:

* Delay webpacker installation to a point where `cucumber-rails` is properly setup in the dummy application `Gemfile`. Otherwise we get warnings like this:

>WARNING: Cucumber-rails required outside of env.rb.  The rest of loading is being deferred
until env.rb is called. To avoid this warning, move 'gem 'cucumber-rails', require: false'under only group :test in your Gemfile. If already in the :test group, be sure you arespecifying 'require: false'.

and also a hard error about `rails` generator not finding the `webpacker:install` task. Honestly I'm not fully sure why it's happening, but we're running `webpacker:install` later on, so it's unnecessary to include `webpacker` on the initial dummy application generation.

* Increase aruba's commands timeout, since `sassc` compilation is very slow, and it comes by default on Rails 6.

* Bump the nodejs version being used to a version compatible with webpacker. Otherwise we get errors like this:

> Webpacker requires Node.js >= 8.16.0 and you are using 8.12.0
> Please upgrade Node.js https://nodejs.org/en/download/

## Motivation and Context

To have a green CI.

## How Has This Been Tested?

In CI.

## Types of changes

## Checklist:

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,6 @@
 language: ruby
 
-dist: xenial
+dist: bionic
 
 rvm:
   - 2.3
--- a/features/support/aruba.rb
+++ b/features/support/aruba.rb
@@ -1,5 +1,5 @@
 # frozen_string_literal: true
 
 Aruba.configure do |config|
-  config.exit_timeout = 120
+  config.exit_timeout = 360
 end
--- a/features/support/cucumber_rails_helper.rb
+++ b/features/support/cucumber_rails_helper.rb
@@ -37,7 +37,9 @@ module CucumberRailsHelper
 
   def run_rails_new_command(options)
     options[:name] ||= 'test_app'
-    run_command "bundle exec rails new #{options[:name]} --skip-bundle --skip-test-unit --skip-spring --skip-bootsnap #{options[:args]}"
+    flags = '--skip-bundle --skip-test-unit --skip-spring --skip-bootsnap'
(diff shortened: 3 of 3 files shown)


deivid-rodriguez:
This is ready for a review!

deivid-rodriguez:
Thanks for your reviews!

I made one last minute change to fix the "Nodejs version is too old" issue. Instead of all the workarounds to upgrade the nodejs version manually, I updated `TravisCI` configuration to use `bionic` which seems to provide a high enough version of `nodejs` by default.

deivid-rodriguez:
I will merge this after the CI :+1:.

deivid-rodriguez:
Going in!

Issue #455, opened 2020-01-08
cucumber-rails: Ruby 2.7 readiness issue


olleolleolle:
## Summary

We want to be compatible with current Ruby versions. 2.7 is not yet in the matrix. It could be.

## Expected Behavior

- Have 2.7 in CI matrix
- No Ruby 2.7 warnings from the test suite
- All tests pass (or, no tests fail due to 2.7)

## Current Behavior

- Does not have Ruby 2.7 in the test matrix

## Possible Solution

- Extend the matrix.

## Context & Motivation

Rails is moving forwards with their warnings-removal work for 2.7. This is a Rails plugin. Would be nice to move in the same direction.

luke-hill:
I would advise that this is `hard-blocked` by the issues in cucumber / cucumber-core repos.

https://github.com/cucumber/cucumber-ruby-core/issues/187
https://github.com/cucumber/cucumber/issues/843

olleolleolle:
@luke-hill Both of these issues are closed.

DannyBen:
Since the `cucumber-ruby` gem is in [4.0 Release Candidate state for over a year](https://github.com/cucumber/cucumber-ruby/releases), may I suggest to consider at least releasing a `cucumber-rails` RC version, with the latest and greatest `cucumber`(RC) and `cucumber-core`?

This lag is beginning to have an impact on productivity, I guess not only for me.

luke-hill:
@DannyBen This issue is now hard blocked on #466 and I suggest once #466 is released we can probably close this as no longer relevant / fixed.

as for getting cucumber ruby 4 support in cucumber rails, This was done recently in the Open Source work. However it's anticipated some small issues might come crawling out.

olleolleolle:
@mvz You've gotten #466 merged - and I just checked, no warnings are showing in the Job

- 5.2 https://travis-ci.org/github/cucumber/cucumber-rails/jobs/697534322
- 6.0 https://travis-ci.org/github/cucumber/cucumber-rails/jobs/697534321

Yay!

PR #451, opened 2019-12-13
Update rubocop

mvz:
## Summary

Bump rubocop dependency and update config file with new cop namespace.

## Details

Rubocop 0.78.0 changed the namespace for the LineLength cop.

## Motivation and Context

Keeps things up-to-date.

## How Has This Been Tested?

I ran rubocop.

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue).
- [ ] New feature (non-breaking change which adds functionality).
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected).

## Checklist:

- [ ] I've added tests for my code.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.

PR #466, opened 2020-06-06
Support for Ruby 2.4 through 2.7

mvz:
## Summary

Update the set of tested and supported Rubies to 2.4 - 2.7

## Details

- Bump minimum version in gemspec and rubocop config
- Stop building on 2.3 on Travis
- Start building on 2.7 on Travis
- Update rubocop and aruba dependencies, which also support Ruby 2.4+

## Motivation and Context

This relates to #455 and lowers the maintenance burden.

## How Has This Been Tested?

Travis will have to check on all supported Rubies.

## Types of changes

Not sure.

## Checklist:

Not sure yet either.

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -20,6 +20,34 @@ Metrics/BlockLength:
 Style/NumericLiteralPrefix:
   EnforcedOctalStyle: zero_only
 
+# Enable new cops
+Layout/EmptyLinesAroundAttributeAccessor:
+  Enabled: true
+Layout/SpaceAroundMethodCallOperator:
+  Enabled: true
+Lint/DeprecatedOpenSSLConstant:
+  Enabled: true
+Lint/MixedRegexpCaptureTypes:
+  Enabled: true
+Lint/RaiseException:
+  Enabled: true
+Lint/StructNewOverride:
+  Enabled: true
+Style/ExponentialNotation:
+  Enabled: true
+Style/HashEachMethods:
+  Enabled: true
+Style/HashTransformKeys:
+  Enabled: true
+Style/HashTransformValues:
+  Enabled: true
+Style/RedundantRegexpCharacterClass:
+  Enabled: true
+Style/RedundantRegexpEscape:
+  Enabled: true
+Style/SlicingWithRange:
+  Enabled: true
+
 ## Cucumber Repo styles (Across implementations) ##
 
 Layout/LineLength:
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -3,7 +3,7 @@ require:
   - rubocop-performance
 
 AllCops:
-  TargetRubyVersion: 2.3
+  TargetRubyVersion: 2.4
   Exclude:
     # These are auto-generated from a load of features that use aruba
     - 'tmp/**/*'
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,7 +3,6 @@ language: ruby
 dist: bionic
 
 rvm:
-  - 2.3
   - 2.4
   - 2.5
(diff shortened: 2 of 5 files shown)


.travis.yml
```diff
@@ -33,10 +33,6 @@ matrix:
   exclude:
   # Only test Rails6 on supported rubies
   # Don't test lowest rails support on maintained rubies
-    - rvm: 2.3
-      gemfile: gemfiles/rails_6_0.gemfile
-    - rvm: 2.4
-      gemfile: gemfiles/rails_4_2.gemfile
     - rvm: 2.4
       gemfile: gemfiles/rails_6_0.gemfile
     - rvm: 2.5
```
mathieujobin, line 38: I think tests should run for ruby 2.5 and rails 4.2 combo as well

mvz: I agree but tried to go keep with the spirit of the set of exclusions that was already there.

Perhaps @luke-hill, who created this set in the first place, can explain the reasoning for not testing 4.2 on Ruby 2.5?

luke-hill: At the time we originally drafted the 2.5 addition to travis (Or at least first started culling the travis releases), we made a conscious decision that the "newest" ruby at the time wouldn't support the oldest rails we supported. Because that Max/Min combo was perceived to be not desirable.

Also 2.5 (For another reason), was a Rails6 enforcement. So when we introduced rails6, it felt like most people on 2.5 would go straight to rails6 for big gains.

.travis.yml
```diff
@@ -33,10 +33,6 @@ matrix:
   exclude:
   # Only test Rails6 on supported rubies
   # Don't test lowest rails support on maintained rubies
-    - rvm: 2.3
-      gemfile: gemfiles/rails_6_0.gemfile
-    - rvm: 2.4
```
luke-hill, line 38: So by the logic below, I think this should stay red.

.travis.yml
mvz: Right, so Rails 4.2 would only be tested with Ruby 2.4. This makes sense to me.

I'm going to mull over the possibility of inverting the logic here, because with so many exclusions I find it hard to have a clear overview of what will be built. I don't know if that's possible with Travis' config, but I'll see.

luke-hill: My philosophy here is if you can justify the change that's fine. I'm not gonna rule with an iron fist here.

By default (With simple maths), we have 5*5 tested permutations. And I think Rails6 rules out the top2/3. So realistically by default we have like 22 permutations.

We need to get that down a little bit, but then I suppose it's also about volume. If we're not really firing many PR's at it, I'm happy to open it up a little bit, I just wouldn't want something previously excluded to be "reincluded" so to speak.

luke-hill:
@mvz can we move this PR into ready for review? As someone else has come forward to do some follow-up work to this.

.rubocop.yml
```diff
@@ -20,6 +20,34 @@ Metrics/BlockLength:
 Style/NumericLiteralPrefix:
   EnforcedOctalStyle: zero_only
 
+# Enable new cops
```
luke-hill, line 23: We can use a new style here called
```
AllCops:
  NewCops: enable
```

Nested underneath the above.

deivid-rodriguez: For what is worth, I prefer explicitly enabling cops that we're interested in (even if it's all of them), rather than blindly accepting every new from rubocop.

mvz:
> @mvz can we move this PR into ready for review? As someone else has come forward to do some follow-up work to this.

Done! I'll also rebase this now that #451 has been merged.

Issue #459, opened 2020-02-04
DatabaseCleaner 1.8.x deprecates DatabaseCleaner.connections, which cucumber-rails uses


christiannelson:
## Summary

`DatabaseCleaner.connections` has been deprecated in the 1.8.x versions of DatabaseCleaner. The deprecation is triggered from cucumber rails on this line: https://github.com/cucumber/cucumber-rails/blob/a33bf05bf060339091af54a7f6019d48eba31b5d/lib/cucumber/rails/database.rb#L79

## Expected Behavior

No deprecation warning when using the latest DatabaseCleaner versions.

## Current Behavior

~~~
Calling `DatabaseCleaner.connections` is deprecated, and will be removed in database_cleaner 2.0. Use `DatabaseCleaner.cleaners`, instead.
~~~

## Possible Solution

Call `DatabaseCleaner.connections` on older versions of DatabaseCleaner and `.cleaners` on newer ones. I'm not 100% certain when cleaners was introduced though.

## Context & Motivation

We like to have warning-free test runs. :) 

## Your Environment

* Version used: `cucumber-rails (2.0.0)`
* Operating System and version: `macOS 10.15.3`
* Link to your project: N/A

luke-hill:
Whilst everything in this report is fair, it's probably worth noting we still officially support rails 4.2 (And I'm not sure what version of Cleaner is tied to what version of rails).

Putting in some branching logic is quite easy though, so this ticket will be worked on, we just need to cover our bases.

mathieujobin:
I'm in the progress of upgrading a rails 4.2 app and I just bumped database_cleaner to 1.8 without any problems.

Issue #406, opened 2019-03-13
The functionality of database cleaner has been integrated to rails since 5.1


LionelLindt:
## Summary

The functionality of the `database-cleaner` gem has been integrated to rails since `5.1`. I therefore think, it shouldn't be used by `cucumber`.

## Expected Behavior

I expect `cucumber` to work without the `database-cleaner` gem, like rails system tests do too since version `5.1`.

## Current Behavior

When I run a `cucumber` feature test without `database-cleaner`, some test won't work.

## Context & Motivation

I'd like to use `cucumber` without the `database-cleaner` gem because it is an additional dependency and it is not needed since rails `5.1`

luke-hill:
It's a valid point. But technically at the moment Rails 4.2 is our lowest supported version. So we'd need to keep supporting it until Rails 5.1 becomes our lowest supported version (I've no idea when this is likely to be, but conservative estimates would be at least another year)

coorasse:
I find this a relevant point. Why can't we have support for Rails 5.1 System tests and keep requiring database_cleaner for 4.2? Is there an impediment for that?

luke-hill:
We could do. But does it provide value....

If someone raised a PR I could review it, but it's not something I'd work on at the moment

coorasse:
I think it would provide quite an added value: remove the dependency from database_cleaner. The system tests provided by Rails work with transactions (and not truncation or deletion of the database), therefore are also faster.
I understand this is not something you might be willing to do, but if you consider this a valid point and you are willing to accept PRs maybe someone else can take a look into it 👍

luke-hill:
My knowledge of rails is weak, but from what the OP originally said is that, this functionality is provided by default from `5.1`

What I'm saying is `5.1` is not the lowest version we support, it's `4.2`, so we would need some conditional. Which I don't think makes sense. We should just wait until `5.1` is the lowest version and then delete the dependency?

luke-hill:
@coorasse is this something you were wanting to work on, or just something you were asking about? Just wondering whether to keep this issue open or not. If it adds value to keep it open I'm not against it. Just the timeline for removing rails 4.2 and 5.0 is essentially non-existent.

coorasse:
[...]

PR #467, opened 2020-06-06
Compatibility with database_cleaner v2 adapter gems

botandrose:
## Summary

Hi folks! This PR aims to resolve some incompatibilities with cucumber-rails and the upcoming major v2 release of database_cleaner.

## Details
In order to get access to the database_cleaner configuration in a ORM-agnostic way, we first try to require `database_cleaner/core` (which is only available in the database_cleaner v2 family of gems), then fall back to requiring `database_cleaner` for folks still on v1, finally falling back to silence if neither is available, since database_cleaner is an optional dependency.

## Motivation and Context

Last week, we cut beta versions of the next major version of database_cleaner. The main change here is that all of the ORM adapters have been split off into their own gems, e.g. database_cleaner-active_record, database_cleaner-redis, etc. Understanding that the majority of database_cleaner's current users are using it with Rails/ActiveRecord (and this gem, too), we chose to make the database_cleaner v2.0.0 gem a metagem that depends on database_cleaner-active_record, and simply loads that when `require "database_cleaner"` is run. This means that the vast majority of our users won't have to touch a thing to upgrade from database_cleaner v1 to v2. 

However, not everyone is only using ActiveRecord. There are those who can't get away with just leaving `gem "database_cleaner"` as-is in their Gemfile, and they have a bit more work to do. For example, let's say you're using Rails with Sequel and Redis, and want to upgrade to database_cleaner v2.0.0. No big deal, you just replace `gem "database_cleaner"` with `gem "database_cleaner-sequel"` and `gem "database_cleaner-redis"` and you're good to go. However, in this situation, the current implementation of cucumber-rails' database_cleaner integration will silently fail.

The underlying issue is that none of the ORM adapters have a `lib/database_cleaner.rb` file (only the database_cleaner metagem has it, for backwards compatibility with the common case), so the `require "database_cleaner"` call in the "lib/cucumber/rails/hooks/database_cleaner.rb" will silently fail since `LoadError` is being caught and discarded, the only visible clue being that the database state is now persisting between test cases. Personally, I consider this issue to be a blocker for releasing final v2.0.0 versions of the database_cleaner family of gems.

## How Has This Been Tested?

The first commit in this PR adds a new cucumber feature that fails, exposing this issue. It does this by simply adding `gem "database_cleaner-active_record"` to the Gemfile, instead of `gem "database_cleaner"`. This was the simplest way to expose the issue without adding a dev dependency on another ORM. This is also a valid and supported way to use database_cleaner with ActiveRecord, BTW.

## Types of changes

## Checklist:

This change should be transparent to consumers of the library.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,7 +5,7 @@ on how to contribute to Cucumber.
 
 ### New Features
 
- * 
+ *
 
 ### Changed
 
@@ -13,7 +13,8 @@ on how to contribute to Cucumber.
 
 ### Fixed
 
- *
+ * Database cleaning no longer silently fails when using database_cleaner v2 adapter gems
+   ([#467](https://github.com/cucumber/cucumber-rails/pull/467) [botandrose])
 
 ## [v2.1.0](https://github.com/cucumber/cucumber-rails/compare/v2.0.0...v2.1.0) (2020-06-15)
 
--- a/features/support/cucumber_rails_helper.rb
+++ b/features/support/cucumber_rails_helper.rb
@@ -13,6 +13,7 @@ module CucumberRailsHelper
     add_gem 'selenium-webdriver', '~> 3.11', group: :test
     add_gem 'rspec-expectations', '~> 3.7', group: :test
     add_gem 'database_cleaner', '>= 1.8.0', group: :test unless options.include?(:no_database_cleaner)
+    add_gem 'database_cleaner-active_record', '>= 2.0.0.beta2', group: :test if options.include?(:database_cleaner_active_record)
     add_gem 'factory_bot', '>= 3.2', group: :test unless options.include?(:no_factory_bot)
 
     run_command_and_stop 'bundle install'
--- a/lib/cucumber/rails/hooks/database_cleaner.rb
+++ b/lib/cucumber/rails/hooks/database_cleaner.rb
@@ -1,8 +1,16 @@
 # frozen_string_literal: true
 
 begin
-  require 'database_cleaner'
+  require 'database_cleaner/core'
+rescue LoadError
+  begin
+    require 'database_cleaner'
+  rescue LoadError
+    Cucumber.logger.debug('neither database_cleaner v1 or v2 present')
+  end
+end
 
+if defined?(DatabaseCleaner)
   Before('not @no-database-cleaner') do
     DatabaseCleaner.start if Cucumber::Rails::Database.autorun_database_cleaner
   end
@@ -10,6 +18,4 @@ begin
   After('not @no-database-cleaner') do
     DatabaseCleaner.clean if Cucumber::Rails::Database.autorun_database_cleaner
   end
-rescue LoadError
-  Cucumber.logger.debug('database_cleaner gem not present.')
 end
--- a/features/database_cleaner.feature
+++ b/features/database_cleaner.feature
@@ -1,7 +1,50 @@
 Feature: Database Cleaner
 
-  Scenario: Create records in background
+  Scenario: Create records in background with database_cleaner
     Given I have created a new Rails app and installed cucumber-rails
+    When I write to "features/widgets.feature" with:
+      """
+      Feature: Create widgets
+        Background: 2 initial widgets
+          Given I have 2 widgets
+
+        Scenario: Add 3 widgets
+          When I create 3 more widgets
+          Then I should have 5 widgets
+
+        Scenario: Add 7 widgets
+          When I create 7 more widgets
+          Then I should have 9 widgets
+      """
+    And I run `rails generate model widget name:string`
+    And I write to "features/step_definitions/widget_steps.rb" with:
+      """
+      Given('I have {int} widgets') do |number|
+        number.times do |i|
+          Widget.create! name: "Widget #{Widget.count + i}"
+        end
+      end
+
+      When('I create {int} more widgets') do |number|
+        number.times do |i|
+          Widget.create! name: "Widget #{Widget.count + i}"
+        end
+      end
+
+      Then('I should have {int} widgets') do |number|
+        expect(Widget.count).to eq(number)
+      end
+      """
+    And I run `bundle exec rake db:migrate`
+    And I run `bundle exec rake cucumber`
+    Then the feature run should pass with:
+      """
+      2 scenarios (2 passed)
+      6 steps (6 passed)
+      """
+
+  Scenario: Create records in background with database_cleaner-active_record
+    Given I have created a new Rails app and installed cucumber-rails with database_cleaner-active_record
     When I write to "features/widgets.feature" with:
       """
       Feature: Create widgets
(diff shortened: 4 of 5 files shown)


botandrose:
Okay, tests have all completed for the failing test commit. I've checked the output for all the runners, and they are all failing as expected, with the following:

```
Using the default profile...
Feature: Create widgets
  Background: 2 initial widgets # features/widgets.feature:2
    Given I have 2 widgets      # features/step_definitions/widget_steps.rb:1
  Scenario: Add 3 widgets        # features/widgets.feature:5
    When I create 3 more widgets # features/step_definitions/widget_steps.rb:7
    Then I should have 5 widgets # features/step_definitions/widget_steps.rb:13
  Scenario: Add 7 widgets        # features/widgets.feature:9
    When I create 7 more widgets # features/step_definitions/widget_steps.rb:7
    Then I should have 9 widgets # features/step_definitions/widget_steps.rb:13
      expected: 9
           got: 14
      (compared using ==)
       (RSpec::Expectations::ExpectationNotMetError)
      ./features/step_definitions/widget_steps.rb:14:in `"I should have {int} widgets"'
      features/widgets.feature:11:in `I should have 9 widgets'
Failing Scenarios:
cucumber features/widgets.feature:9 # Scenario: Add 7 widgets
2 scenarios (1 failed, 1 passed)
6 steps (1 failed, 5 passed)
0m0.155s
..................................................................
(::) failed steps (::)
expected that command "bundle exec rake cucumber" has exit status of "0", but has "1". (RSpec::Expectations::ExpectationNotMetError)
./features/step_definitions/cucumber_rails_steps.rb:72:in `"the feature run should pass with:"'
features/database_cleaner.feature:83:in `the feature run should pass with:'
Failing Scenarios:
cucumber features/database_cleaner.feature:46 # Scenario: Create records in background with database_cleaner-active_record
```

botandrose:
Okay, we're green! Ready for review. What do y'all think?

luke-hill:
Hi there. So provisionally I'm ok with this. I'm trying to find a commit, but maybe you can help.

Someone mentioned that with a new-er version of Rails (I can't remember if it was 5.2 or 6.0), database cleaner was no longer required and sort of auto included. But I can't remember the details as it was a long time ago. I remember the thought process here was we were going to *remove* database cleaner at some point.

If I'm wrong that's fine. You may be able to piece things together better for me. I'll try have a dig around. For now can we just hold off. But if everything's ok I'm happy with the approach here, just need to confirm my original thought process was wrong.

luke-hill:
https://github.com/cucumber/cucumber-rails/issues/406 &
https://github.com/cucumber/cucumber-rails/issues/459

BrianHawley:
It was in Rails 5.1, and it's not strictly that database_cleaner isn't required. It was more that they made it possible to run tests in a mode where all threads share a database connection, and this mode makes it possible to have transaction-mode cleanup actually work in multi-threading environments like capybara.

However, that mode does change the behavior of the app: all threads are sharing one database connection. Rails doesn't have built-in support for any of the other database_cleaner cleanup modes, which you'd still need to use when testing certain kinds of concurrency issues.

Fortunately the new test concurrency mode works with database_cleaner's transactional cleanup mode, so even if you have it installed, it can still be used.

luke-hill:
Ok so in essence, adding this new additional permutation won't alter the use case for people who no longer want to use database cleaner in rails 5.1+ ?

lib/cucumber/rails/hooks/database_cleaner.rb
```diff
@@ -1,15 +1,21 @@
 # frozen_string_literal: true
 
 begin
-  require 'database_cleaner'
+  require 'database_cleaner/core'
+rescue LoadError
+  begin
+    require 'database_cleaner'
+  rescue LoadError
+    Cucumber.logger.debug('database_cleaner not present.')
```
luke-hill, line 9: Probably worth hinting that this has checked both gems right?

luke-hill:
Also when you come back to this, could you rebase with latest master, and add a changelog entry at the top in unreleased

lib/cucumber/rails/hooks/database_cleaner.rb
botandrose: Sure! Are you imagining something like "neither database_cleaner v1 or v2 present"?

botandrose:
@luke-hill Rebased and changelog updated. Sorry it took so long to get back to this. Last week was crazy IRL. Looks like I missed the v2.1 release train, too. Crap!

luke-hill:
Trying to see if the failure was a genuine timeout or something else (Fairly sure travis just had a moment).

CHANGELOG.md
```diff
@@ -5,15 +5,16 @@ on how to contribute to Cucumber.
...
+ *
 
 ### Changed
 
  *
 
 ### Fixed
 
- *
+ * Database cleaning silently fails when using database_cleaner v2 adapter gems
```
luke-hill, line 16: should this say something like "no longer fails", or have I misunderstood this?

features/support/cucumber_rails_helper.rb
```diff
@@ -13,6 +13,7 @@ def install_cucumber_rails(*options)
     add_gem 'selenium-webdriver', '~> 3.11', group: :test
     add_gem 'rspec-expectations', '~> 3.7', group: :test
     add_gem 'database_cleaner', '>= 1.8.0', group: :test unless options.include?(:no_database_cleaner)
+    add_gem 'database_cleaner-active_record', '>= 2.0.0.beta2', group: :test if options.include?(:database_cleaner_active_record)
```
luke-hill, line 16: Is this beta copy the best we have (This helper is used for everyone who uses the gem). Do we have any idea how close to a stable 2.x revision we are?

lib/cucumber/rails/hooks/database_cleaner.rb
luke-hill: Something like that yeh

CHANGELOG.md
botandrose: Ah, I think I see. I was describing the bug, and you would prefer that I describe the bugfix?

features/support/cucumber_rails_helper.rb
botandrose: Yes, this issue is the last blocker for releasing v2.0.0 final. So, as soon as this PR is merged, and out in the wild in a cucumber-rails gem release, I will cut the final v2.0.0 release of database_cleaner and its family of adapter gems.

botandrose:
@luke-hill I have just pushed the requested changes the changelog message and the warning message.

features/support/cucumber_rails_helper.rb
botandrose: If we could coordinate a simultaneous release, we could change this to `>= 2.0.0`

deivid-rodriguez: I'm confused, how is this helper used by everyone using this gem? This file is shipped with the gem, but its folder (`features`) is not exposed as a $LOAD_PATH directory. I thought these were just internal helpers for testing `cucumber-rails`.

aslakhellesoy:
Hi @botandrose,

Aslak Hellesøy
Creator of Cucumber

PR #482, opened 2020-09-08
Feature/updated support

luke-hill:
## Summary

Updated support a little for gem dependencies (Add notes about all core deps)

## Details

As written above

## Motivation and Context

It removes the amibiguity about certain gem restrictions, slowly brings us more in line

## How Has This Been Tested?

CI

## Types of changes

## Checklist:

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -32,10 +32,11 @@ matrix:
   exclude:
   # Only test latest Rails (Of each major), on actively supported Ruby Versions
   # Only test latest Ruby with latest Rails versions (Of each major)
-  # 2.4 -> Not 5.2/6.0
-  # 2.5 -> Anything
-  # 2.6 -> Probably should be using 5.2+
-  # 2.7 -> Probably should be using 5.2+
+  # 2.4  -> Not 5.2/6.0
+  # 2.5  -> Anything
+  # 2.6  -> Users of this should be using Rails 5.2+
+  # 2.7  -> Users of this should be using Rails 5.2+
+  # head -> Users of this should be using Rails 5.2+
     - rvm: 2.4
       gemfile: gemfiles/rails_5_2.gemfile
     - rvm: 2.4
--- a/.travis.yml
+++ b/.travis.yml
@@ -30,8 +30,14 @@ matrix:
   allow_failures:
     - rvm: ruby-head
   exclude:
-  # Only test Rails6 on supported rubies
-  # Don't test lowest rails support on maintained rubies
+  # Only test latest Rails (Of each major), on actively supported Ruby Versions
+  # Only test latest Ruby with latest Rails versions (Of each major)
+  # 2.4 -> Not 5.2/6.0
+  # 2.5 -> Anything
+  # 2.6 -> Probably should be using 5.2+
+  # 2.7 -> Probably should be using 5.2+
+    - rvm: 2.4
+      gemfile: gemfiles/rails_5_2.gemfile
     - rvm: 2.4
       gemfile: gemfiles/rails_6_0.gemfile
     - rvm: 2.6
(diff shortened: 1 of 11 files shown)


LICENSE
```diff
@@ -1,6 +1,6 @@
 The MIT License
 
-Copyright (c) 2008-2019 Aslak Hellesøy and the Cucumber Team.
+Copyright (c) 2008+ Aslak Hellesøy and the Cucumber Team.
```
mvz, line 3: I don't think you can put an open-ended range here. It's a bummer but we'll have to just update the year every time there's a new release in a new year.

cucumber-rails.gemspec
```diff
@@ -22,11 +22,11 @@ Gem::Specification.new do |s|
...
-  s.add_runtime_dependency('capybara', ['>= 2.12', '< 4'])
-  s.add_runtime_dependency('cucumber', ['>= 3.0.2', '< 5'])
-  s.add_runtime_dependency('mime-types', ['>= 2.0', '< 4'])
-  s.add_runtime_dependency('nokogiri', '~> 1.8')
-  s.add_runtime_dependency('rails', ['>= 5.0', '< 7'])
+  s.add_runtime_dependency('capybara', ['>= 2.18', '< 4']) # We support legacy capybara (But only the last 2.x)
+  s.add_runtime_dependency('cucumber', ['>= 3.0.2', '< 6']) # Support cucumber in the 3.x / 4.x / 5.x revisions
+  s.add_runtime_dependency('mime-types', ['~> 3.2']) # Only support the latest major (3+ years old)
+  s.add_runtime_dependency('nokogiri', '~> 1.8') # Only support the latest major (3+ years old)
+  s.add_runtime_dependency('rails', ['>= 5.0', '< 7']) # We support any version of Rails in the 5.x and 6.x series
```
mvz, line 29: Maybe it's clearer if we put one comment at the top here stating we support the last major version older than three years, or the version from the oldest support release of Rails. In any case, stating the general rule at the top would be nice.

If we want to support capybara 2 we should have at least one Appraisal that includes it.

.travis.yml
```diff
@@ -30,8 +30,15 @@ matrix:
...
     - rvm: ruby-head
   exclude:
-  # Only test Rails6 on supported rubies
-  # Don't test lowest rails support on maintained rubies
+  # Only test latest Rails (Of each major), on actively supported Ruby Versions
+  # Only test latest Ruby with latest Rails versions (Of each major)
+  # 2.4  -> Not 5.2/6.0
+  # 2.5  -> Anything
+  # 2.6  -> Users of this should be using Rails 5.2+
+  # 2.7  -> Users of this should be using Rails 5.2+
```
mvz, line 38: :+1:

mvz (commented):
Hi @luke-hill I just have a couple of remarks.

As an aside, when do you think we can drop Rails 5.0 and 5.1? I think it's about time and would clean things up even more.

LICENSE
luke-hill: I've seen other projects do it? Is there a reason we can't?

cucumber-rails.gemspec
luke-hill: These comments were added from a suggestion by Aslak that we document gems that don't use our auto-update everything script.

Each gem support is different purely because we've not got a huge consumer base and I guess I'm trying to update them slowly.

Could state the general rule at top, happy to do that.

Agree we should probably have a capybara 2 use-case if we're still supporting it (Which we need to do for now).

LICENSE
mvz: You need to document the last copyright year so people know when copyright ends. But IANAL.

mvz (approved):
Looks good!

PR #496, opened 2021-03-26
Reduce gem package size

orien:
## Summary

Reduce the size of the gem by removing unneeded files from the package.

## Details

Reduce the gem size from **40K** to **27K**, a saving of **33%**!

<details>
<summary><b>contents diff</b> (click me)</summary>

```diff
- Appraisals
- Gemfile
- Rakefile
- bin/install_geckodriver.sh
- bin/install_webpacker.sh
- config/cucumber.yml
- cucumber-rails.gemspec
- dev_tasks/cucumber.rake
- dev_tasks/rspec.rake
- dev_tasks/yard.rake
- dev_tasks/yard/default/layout/html/bubble_32x32.png
- dev_tasks/yard/default/layout/html/footer.erb
- dev_tasks/yard/default/layout/html/index.erb
- dev_tasks/yard/default/layout/html/layout.erb
- dev_tasks/yard/default/layout/html/logo.erb
- dev_tasks/yard/default/layout/html/setup.rb
- features/allow_rescue.feature
- features/annotations.feature
- features/capybara_javascript_drivers.feature
- features/choose_javascript_database_strategy.feature
- features/configuration.feature
- features/database_cleaner.feature
- features/disable_automatic_database_cleaning.feature
- features/emulate_javascript.feature
- features/install_cucumber_rails.feature
- features/no_database.feature
- features/raising_errors.feature
- features/rerun_profile.feature
- features/rest_api.feature
- features/step_definitions/cucumber_rails_steps.rb
- features/support/aruba.rb
- features/support/cucumber_rails_helper.rb
- features/support/env.rb
- features/support/hooks.rb
- gemfiles/rails_5_0.gemfile
- gemfiles/rails_5_1.gemfile
- gemfiles/rails_5_2.gemfile
- gemfiles/rails_6_0.gemfile
- spec/cucumber/rails/database_spec.rb
- spec/generators/cucumber/install_generator_spec.rb
- spec/spec_helper.rb
```

</details>

## Motivation and Context

Waste not, want not: let's not consume disk space or network bandwidth needlessly.

## Types of changes

## Checklist:

### DIFF
--- a/cucumber-rails.gemspec
+++ b/cucumber-rails.gemspec
@@ -1,6 +1,5 @@
 # frozen_string_literal: true
 
-require 'rake/file_list'
 $LOAD_PATH.unshift File.expand_path('lib', __dir__)
 
 Gem::Specification.new do |s|
@@ -47,7 +46,6 @@ Gem::Specification.new do |s|
 
   s.required_ruby_version = '>= 2.4.0'
   s.rubygems_version = '>= 1.6.1'
-  s.files            = Rake::FileList['**/*'].exclude(*File.read('.gitignore').split)
-  s.test_files       = Dir['{spec,features}/**/*']
   s.require_path     = 'lib'
+  s.files            = Dir['lib/**/*', 'CHANGELOG.md', 'CONTRIBUTING.md', 'LICENSE', 'README.md']
 end


luke-hill (commented):
This was something introduced by someone working on the debian side of things. So I'd like to hear from him.

@utkarsh2102 thoughts on this change? I know you introduced rake file lists a bit ago. But in small gems I maintain. I just use Dir

orien:
Requiring `rack/file_lists` is problematic, because this gemspec is specifying the development dependency on `rake`. However the `require` instruction is made before the dependency has come into effect. A chicken and egg problem. This prevents projects from including `cucumber-rails` in their Gemfiles like so:

```ruby
gem 'cucumber-rails', git: 'https://github.com/cucumber/cucumber-rails.git'
```

```
[!] There was an error while loading `cucumber-rails.gemspec`: cannot load such file -- rake/file_list
```

At least on Ruby 2.6.6, with Bundler 2.2.15.

cucumber-rails.gemspec
```diff
@@ -47,7 +46,10 @@ Gem::Specification.new do |s|
...
   s.required_ruby_version = '>= 2.4.0'
   s.rubygems_version = '>= 1.6.1'
-  s.files            = Rake::FileList['**/*'].exclude(*File.read('.gitignore').split)
-  s.test_files       = Dir['{spec,features}/**/*']
   s.require_path     = 'lib'
+  s.files            = Dir.chdir(File.expand_path(__dir__)) do
+                         `git ls-files -z`.split("\x0").select do |file|
+                           file.match(%r{^(lib/|CHANGELOG|CONTRIBUTING|LICENSE|README)})
+                         end
+                       end
```
utkarsh2102, line 54: Thanks, however, this change was done explicitly to not use `git` in gemspec, which is problematic for downstream maintainers and even in general; cf: https://docs.rubocop.org/rubocop-packaging/cops_packaging.html#packaginggemspecgit

This is not only Debian-specific but applies to most of the downstream maintainers.

That said, I'd be happy to shift from `Rake::FileList` to `Dir` (which is [what I personally use](https://github.com/utkarsh2102/rubocop-packaging/blob/master/rubocop-packaging.gemspec#L21)). So maybe more like `Dir["lib/**/*", "CHANGELOG", ...]` should be a good replacement here.

What do you think?

utkarsh2102 (changes requested):

cucumber-rails.gemspec
orien: I'm happy to remove references to `git`. 👍

aslakhellesoy: I agree - we should not depend on `git` when building the gemspec.

luke-hill: Seems we have a consensus :D

orien: Added in 89b14cf. The resulting gem file has identical contents as the `git` version.

luke-hill (approved):

cucumber-rails.gemspec
```diff
@@ -47,7 +46,6 @@ Gem::Specification.new do |s|
 
   s.required_ruby_version = '>= 2.4.0'
   s.rubygems_version = '>= 1.6.1'
-  s.files            = Rake::FileList['**/*'].exclude(*File.read('.gitignore').split)
-  s.test_files       = Dir['{spec,features}/**/*']
```
luke-hill, line 51: Do we not want refs to these?

utkarsh2102:
@luke-hill, hey, can you check the CI failures (eg: https://travis-ci.org/github/cucumber/cucumber-rails/jobs/764511904) and confirm that this isn't related? If not related, can this be fixed, if trivial, in another PR?

luke-hill:
Checking into this @utkarsh2102 / @orien but I've got other stuff on. This might not get done til Monday now as there's some other issues here regarding the releasing I'm trying to get on top of also.

orien:
There's no rush @luke-hill, take your time. I appreciate your consideration.

cucumber-rails.gemspec
utkarsh2102: I don't think so, maybe? Do you want to have `test_files` defined? I personally do so, but it's okay to not do that, really. So your call 😄

utkarsh2102:
> Checking into this @utkarsh2102 / @orien but I've got other stuff on. This might not get done til Monday now as there's some other issues here regarding the releasing I'm trying to get on top of also.

Oh absolutely! Please take your time and no rush at all! Whilst I am okay with this PR being merged but the CI jobs are all red. If that goes green, I'd love to take a final look at this and then squash & merge. So let me know once you get time to do that. And enjoy your weekend! 🎉

cucumber-rails.gemspec
orien: FYI: Any files specified in `test_files` will be added to the gem package.

Is there an expectation for downstream `cucumber-rails` users to run these tests?

mvz:
I don't like using `Dir` to find files to include in the gem since there's no guard against including whatever happens to be in those directories in the gem. The git trick guards against that, and so does the Rake::FileList construct, to some extent. However, they both introduce dependencies that may not be present in all circumstances.

I've come to prefer using a manifest file that explicitly lists the files to be included. To help maintain that, I've created [rake-manifest](https://github.com/mvz/rake-manifest).

orien:
`rake-manifest` looks good to me. I'm happy to update this PR to use this.

I see you have `aruba` [using this gem](https://github.com/cucumber/aruba/blob/808c26adab5c6d658cf612b0eb5bba3767d7a881/aruba.gemspec#L47), but `cucumber-ruby` is [still using `Dir#[]`](https://github.com/cucumber/cucumber-ruby/blob/master/cucumber.gemspec#L53-L59).

Do we want to consolidate and use `rake-manifest` for all?

mvz:
> Do we want to consolidate and use `rake-manifest` for all?

I'd love to, but the `cucumber-ruby` maintainers will have to decide for themselves.

cucumber-rails.gemspec
utkarsh2102: Most of the downstream maintainers run upstream tests to see if there's no regression or so. So we definitely need tests. That said, we, in Debian, pull from the GitHub source instead of rubygems for this case, so it's okay to not add here. Or if you want to add for consistency purposes, that's okay (in fact, good) as well. So I'm okay with either, really 😅

utkarsh2102:
> I'd love to, but the cucumber-ruby maintainers will have to decide for themselves.

Me likey the idea 😄 
But again, I'll let @luke-hill and @aslakhellesoy have an authoritative say on this!

luke-hill:
If we retain the test files but make the other changes, what benefit to we get? I'd like to see how big the diff is?

If it's a negligible diff. We can merge as is, and keep them, that way there's no real change.

luke-hill (approved):

orien:
Note, I've got slightly different gem sizes now. In my original test the gem built off `master` master was 60K. However when I build now it's 40K. There may have been some extra file added by mistake or something has changed on `master`.  I've updated the PR description to accomodate a 40K baseline.

Interestingly `cucumber-rails` [v2.2.0](https://rubygems.org/gems/cucumber-rails/versions/2.2.0) on rubygems.org is a whopping 134K! I took a look at the contents and it includes a `tmp/aruba/test_app/` directory. This is a full Rails app that gets created during the test suite.

Considering adding the `test_files` specification to the gem spec, do we need to add everything required to run the tests? I assume that'd be:

```ruby
s.test_files = Dir['{spec,features,config}/**/*', 'cucumber-rails.gemspec', 'Gemfile', 'Rakefile']
```

With this configuration the gem weighs in at 36K. A saving of 10%.

<details>
<summary><b>File contents diff</b> (click me)</summary>

```diff
-Appraisals
-bin/install_geckodriver.sh
-bin/install_webpacker.sh
-dev_tasks/cucumber.rake
-dev_tasks/rspec.rake
-dev_tasks/yard.rake
-dev_tasks/yard/default/layout/html/bubble_32x32.png
-dev_tasks/yard/default/layout/html/footer.erb
-dev_tasks/yard/default/layout/html/index.erb
-dev_tasks/yard/default/layout/html/layout.erb
-dev_tasks/yard/default/layout/html/logo.erb
-dev_tasks/yard/default/layout/html/setup.rb
-gemfiles/rails_5_0.gemfile
-gemfiles/rails_5_1.gemfile
-gemfiles/rails_5_2.gemfile
-gemfiles/rails_6_0.gemfile
```

</details>


FYI: I notice that neither [`cucumber`](https://github.com/cucumber/cucumber-ruby/blob/master/cucumber.gemspec) nor [`aruba`](https://github.com/cucumber/aruba/blob/master/aruba.gemspec) specify `test_files` in their respective gemspec.

luke-hill:
That's fine. It also explains the spike in size. test_app 100% should not be in the built gem. It's generated during usage of the gem in dev mode (As is a secondary git repo), but yeh all we want is the stuff in lib as per usual. Plus the other bells and whistles.

luke-hill:
Going to merge this in AS IS, and we can revisit this down the line.

aslakhellesoy:
Hi @orien,

Aslak Hellesøy
Creator of Cucumber

PR #497, opened 2021-03-26
Silence warnings about puts versus log

mvz:
## Summary

Replace puts with log to silence Cucumber warnings.

## Details

This replaces puts with log in the After scenario hook that prints output of failed commands when the scenario fails.

## Motivation and Context

Cucumber's old puts behavior is deprecated. This change silences warnings from Cucumber about this.

## How Has This Been Tested?

I re-ran one scenario that fails on my machine to see if the warnings disappeared.

## Types of changes

## Checklist:

### DIFF
--- a/features/support/hooks.rb
+++ b/features/support/hooks.rb
@@ -2,7 +2,7 @@
 
 After do |scenario|
   if scenario.failed?
-    puts last_command_stopped.stdout
-    puts last_command_stopped.stderr
+    log last_command_stopped.stdout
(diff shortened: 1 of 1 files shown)


luke-hill (approved):

luke-hill:
Lets merge #498 in first which fixes part of the CI issue. There's also some other bits I've been working on. But this can go in before those.

olleolleolle (approved):
Looks good!

PR #499, opened 2021-03-26
Raise useful error if no DatabaseCleaner strategies are loaded

### LINKED ISSUE #491: Breaks with database_cleaner 2.0
## Summary

When upgrading from database_cleaner 1.8 to 2.0, 

## Expected Behavior

cucumber-rails works in combination with database_cleaner 2.0.

## Current Behavior

I get the following error when a `@javascript` scenario is started:
~~~
undefined method `strategy' for nil:NilClass (NoMethodError)
~~~

This seems to be coming from https://github.com/cucumber/cucumber-rails/blob/master/lib/cucumber/rails/database/strategy.rb#L11.

## Possible Solution

I see two options:

- Figure out where database_cleaner 2.0 stores the current strategy, if not in `DatabaseCleaner.cleaners`.
- Use *two* configuration options instead of one, and create `Before` blocks like so:
  ```
  Before '@javascript' do
    DatabaseCleaner.strategy = Cucumber::Rails::Database.javascript_strategy
  end

  Before 'not @javascript' do
    DatabaseCleaner.strategy = Cucumber::Rails::Database.plain_strategy
  end
  ```

The latter option avoids having to use non-documented DatabaseCleaner APIs yet again.

## Steps to Reproduce (for bugs)

1. Have a Rails app with javascript scenarios, cucumber-rails, and database_cleaner 1.8.
2. Upgrade database_cleaner to 2.0
3. Run the cukes

## Context & Motivation

This came up when Dependabot tried to upgrade database_cleaner in my work projects.

## Your Environment

* Version used: 2.2.0
* Operating System and version: Debian, unstable
* Link to your project: N/A

mvz:
## Summary

Raise useful error if no DatabaseCleaner strategies are loaded

## Details

When `#before_js` is called on a Strategy, bail out if DatabaseCleaner.cleaners is empty.

## Motivation and Context

This fixes #491.

## How Has This Been Tested?

I added a spec and made it pass.

## Types of changes

## Checklist:

### DIFF
--- a/spec/cucumber/rails/database_spec.rb
+++ b/spec/cucumber/rails/database_spec.rb
@@ -1,5 +1,6 @@
 # frozen_string_literal: true
 
+require 'database_cleaner'
 require 'cucumber/rails/database/strategy'
 require 'cucumber/rails/database/deletion_strategy'
 require 'cucumber/rails/database/shared_connection_strategy'
@@ -21,6 +22,13 @@ describe Cucumber::Rails::Database do
       expect(strategy).to receive(:before_js)
 
       described_class.before_js
+    end
+
+    it 'raises an error on `before_js` if no DatabaseCleaner cleaners exist' do
+      allow(DatabaseCleaner).to receive(:cleaners).and_return({})
+
+      expect { described_class.before_js }
+        .to raise_error /No DatabaseCleaner strategies found/
     end
   end
 
--- a/spec/cucumber/rails/database_spec.rb
+++ b/spec/cucumber/rails/database_spec.rb
@@ -8,8 +8,9 @@ require 'cucumber/rails/database/truncation_strategy'
 require 'cucumber/rails/database'
 
 describe Cucumber::Rails::Database do
+  let(:strategy) { described_class.instance_eval { @strategy } }
+
   context 'when using a valid pre-determined strategy' do
-    let(:strategy) { described_class.instance_eval { @strategy } }
     before { described_class.javascript_strategy = :truncation }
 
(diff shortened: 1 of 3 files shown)


spec/cucumber/rails/database_spec.rb
```diff
@@ -1,18 +1,15 @@
...
 require 'cucumber/rails/database'
 
 describe Cucumber::Rails::Database do
-  before { allow(strategy_type).to receive(:new).and_return(strategy) }
-
-  let(:strategy) { instance_double(strategy_type, before_js: nil, before_non_js: nil) }
-  let(:strategy_type) { Cucumber::Rails::Database::TruncationStrategy }
-
   context 'when using a valid pre-determined strategy' do
+    let(:strategy) { described_class.instance_eval { @strategy } }
```
luke-hill, line 12: Not sure for the idea for changing the specs around?

mvz: I changed this so the specs will use a real strategy instance instead of an `instance_double`.

luke-hill: I've dug in and think I can spot why it was weird. Just not keen on instance_eval in specs. But I see it's because we don't allow people to access the strategy value from the specification.

In that case it's all fine. You can move up the strategy let from both blocks to avoid duplication, but other than that it's good.

orien (approved):

luke-hill:
I can re-review this by monday/tuesday. If you think it's still relevant (Just been off for a while).

luke-hill (approved):

PR #501, opened 2021-03-29
Quick fixes for tests missed in first fix.

luke-hill:
Fix third test from same issue as first test
Fix 2nd test having issue with new assertion on UTC timezone
Rename some cukes


olleolleolle (approved):
Looks reasonable!

orien (approved):
It's green! 🚀

Issue #510, opened 2021-04-07
`rails notes` does not find TODOs in `features/` directory in Rails 6+


orien:
### Summary

In Rails 6, the [`rails notes`](https://guides.rubyonrails.org/v6.0/command_line.html#rails-notes) feature has [migrated](https://github.com/rails/rails/pull/33220) from a Rake implementation to a Rails Command. The Rake hooks, as supplied via `rails generate cucumber:install` are no-longer triggered when running `rails notes`.

https://github.com/cucumber/cucumber-rails/blob/747a9c4b2017f22b497b8e7201e59f708b1f151a/lib/generators/cucumber/templates/tasks/cucumber.rake.erb#L39-L46

This means the `rails notes` command doesn't find `TODO` notes in the `features` directory.

### Expected Behavior

```cucumber
Given I have created a new Rails 6 app and installed cucumber-rails
When I write to "features/products.feature" with:
  """
  Feature: Products
    Scenario: Test a Product
      # TODO: When I go to the products page
  """
When I run `bundle exec rails notes`
Then it should pass with:
  """
  features/products.feature:
    * [3] [TODO] When I go to the products page
  """
```

### Current Behavior

The command produces no output.

### Environment

* Version used: `cucumber-rails` 2.3.0
* Operating System and version: macOS 11.2.3
* Rails: 6.1.3.1

PR #511, opened 2021-04-07
Configure Rails annotations via environment config (Rails 6+)

orien:
### Summary

Fix `rails notes` support by adding config to `config/environments/development.rb` and `config/environments/test.rb` for Rails 6+ projects..

### Context

As per #510, the config generated via `rails generate cucumber:install` does not support `rails notes` for Rails 6+.

By [changing the annotate feature](https://github.com/cucumber/cucumber-rails/pull/511/commits/4e751e5042ad8d7137795d22d57ed3d32a8883f0) to exercise `rails notes` rather than `rake notes` the problem is [exposed](https://github.com/cucumber/cucumber-rails/runs/2284601677).

### Change

Update the install generator so when `rails generate cucumber:install` is run the existing Rake-based annotation configuration is only generated for projects running versions of Rails pre v6.0. When running Rails v6.0 or greater, add configuration to the Rails `development` and `test` environment configuration files.

#### How Has This Been Tested?

This is covered in both RSpec and Cucumber tests.

### Types of changes

### Checklist


spec/generators/cucumber/install_generator_spec.rb
```diff
@@ -7,7 +7,16 @@
...
-  before { prepare_destination }
+  before do
+    prepare_destination
+    FileUtils.mkdir_p File.join(destination_root, 'config', 'environments')
+    %w[development test].each do |environment|
+      File.write(
+        File.join(destination_root, 'config', 'environments', "#{environment}.rb"),
+        "Rails.application.configure do\nend"
+      )
+    end
```
orien, line 18: Given the generator is now modifying existing configuration files, we'll need to ensure these files exist before exercising the system-under-test in the RSpec test suite.

olleolleolle: Cool, perhaps those new lines could have their own local helper method, with a name that captures your contextual comment?

olleolleolle (commented):
Super-nice addition!

spec/generators/cucumber/install_generator_spec.rb
orien: Excellent idea! Added in [`eca5bad`](https://github.com/cucumber/cucumber-rails/compare/884b75e689f15b7e8e335b397f3e586c00276392..eca5bad975acca459a3bf629888cbb8932ca59e2).

luke-hill:
This explains a lot (I was getting the rake notes test failing locally!)

PR #520, opened 2021-07-21
v2.4.0

luke-hill:
Release v2.4.0


luke-hill:
After this is merged I'll release the 2.4.0 today as it permits users to go from cucumber 5.x to cucumber 7.x whilst using cucumber-rails

aurelien-reeves (approved):
Looks good :)

Issue #521, opened 2021-09-22
handling of @javascript tag is confusing


akostadinov:
Presently handling of scenarios tagged with `@javascript` and without is [hardcoded](https://github.com/cucumber/cucumber-rails/blob/40649dc959fdb9adf9c5bbde3b4076b6cf8a7085/lib/cucumber/rails/hooks/active_record.rb). This is very confusing, see https://github.com/cucumber/cucumber-rails/issues/166.

I see no clear documentation when a scenario is supposed to be tagged `@javascript` and then it shouldn't. I believe that it will be much clearer if tag is named `@shared-connection` as well create ability to people to list tags that should be handled one way or another.

In current implementation if project wants to use `@tag1`, `@tag2` and `@tag3` with shared connection transactional clean-up and `@tag4`, `@tag5` and `@tag6` with non-shared connection truncation clean-up, then there seems to be no way of doing this.

For example if one has this in `support/env.rb`:
```ruby
non_transactional_tags = %w[
  @tag4
  @tag5
  @tag6
].freeze

transactional = non_transactional.map {|t| "not #{t}" }

Before(*transactional) do
  Cucumber::Rails::Database.javascript_strategy = :transaction
  Cucumber::Rails::Database.before_js if Cucumber::Rails::Database.autorun_database_cleaner
end

Before non_transactional.join(' or ') do
  Cucumber::Rails::Database.javascript_strategy = :truncation
  Cucumber::Rails::Database.before_non_js if Cucumber::Rails::Database.autorun_database_cleaner
end

require 'cucumber/rails'
```

This doesn't help because default hooks will call `Cucumber::Rails::Database.before_non_js` for anything not tagged `@javascript` even though we ourselves called `before_js`. And if we move `require 'cucumber/rails'` on top, then database cleaner will already be started so our hooks will be pointless.

aurelien-reeves:
Hello :)

The JavaScript tag related to database strategy is documented here: https://github.com/cucumber/cucumber-rails/issues/521

Does that help?

akostadinov:
I found this feature yesterday and it somehow helped understand what is going on, although I didn't really until I read the database strategy hooks and classes. The bigger issue is that one can't configure this behavior easily depending on project tags. I'm trying to workaround this with a custom strategy class now and will post what I come up with here if this may serve as a better example of what I'm trying to do.

luke-hill:
[...]

PR #522, opened 2021-10-01
null database strategy to disable @javascript handling

akostadinov:
## Summary

Presently there is no nice way to disable `@javascript` tag handling without writing a custom database strategy class. See #521

This PR adds a null database strategy to allow easily doing so by

```
Cucumber::Rails::Database.javascript_strategy = :none
```

I'm planning to add some other enhancements and documentation later to allow custom list of tags but having a null strategy is useful anyway and is a first step.

## Types of changes

## Checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,7 +5,7 @@ on how to contribute to Cucumber.
 
 ### New Features
 
-* 
+* Added `:none` option for `javascript_strategy` which indicates no special handling of `@javascript` tagged tests ([#522](https://github.com/cucumber/cucumber-rails/pull/522) [akostadinov])
 
 ### Changed
 
--- a/features/choose_javascript_database_strategy.feature
+++ b/features/choose_javascript_database_strategy.feature
@@ -20,8 +20,11 @@ Feature: Choose javascript database strategy
   Right now, the default behavior is to use truncation, but you can override this by telling
   cucumber-rails which strategy to use for javascript scenarios.
 
-  The deletion strategy can be quicker for situations where truncation causes locks which
+  The `:deletion` strategy can be quicker for situations where truncation causes locks which
   has been reported by some Oracle users.
+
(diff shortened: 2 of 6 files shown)


aurelien-reeves:
What about naming it "none" instead of "null"?

akostadinov:
@aurelien-reeves , I agree with whatever naming project wants. If the change is generally accepted, I'd go ahead to rename it.

akostadinov:
I renamed to none.

aurelien-reeves (commented):
It looks good to me, well done 👌
But as I am not that much into this project, I prefer to let other contributors to review it :)

aurelien-reeves:
@akostadinov could you add an entry in the changelog? And if you had the opportunity to add some documentation regarding those database strategies, that would be a good bonus to that PR :)

@olleolleolle @luke-hill if no further objection, I may merge this soon

luke-hill (changes requested):
This feels like it's not a None strategy i.e. one isn't there, but a null strategy. So perhaps we should name it NullStrategy to infer that the responses on the js checkers are nil

olleolleolle (approved):
Looks good to me.

akostadinov:
Due to popular demand, I renamed class back to `NullStrategy`. And `:none` as strategy selection makes sense to me to indicate that we use none of the javascript testing strategies. `:null` also makes sense to be consistent with the class name though. Probably `none` reads a little better. I have arguments in favor of both options so I can change to whatever project owners want.

@luke-hill , added changelog and updated doc to clarify `none` and `deletion` are symbols. I have some other modifications ready that can make the feature useful for whatever tags a project decides to use. I think it's better to submit changes in steps and this is the first step. I want to add information to the readme once I have a review and some agreement about the other changes. Is this ok?

akostadinov:
Guys, can we merge in case of no objections?

Issue #523, opened 2021-12-16
Support for Rails 7


erikaxel:
Rails 7 has been released and it would be great if cucumber-rails could support it. 

I did an initial "fix" where I just relaxed the gemspec requirement and added test runners for 7.0 in https://github.com/erikaxel/cucumber-rails/commit/6b225611fad8af12780f3718bd9f9c63327c4462
However that seems to be insufficient since a couple of tests fail. I'm not very familiar with this project so I'm a bit stumped on where to continue. Happy to help if I can, but it would be great with some pointers on where to start. The errors on the CI can be seen here: https://github.com/erikaxel/cucumber-rails/actions/runs/1587333272
 
PS: Also happy to make my initial commit with tests running on 7.0 into a PR if that is helpful.

luke-hill:
You're a bit keen on the gun :) But yes I will get to this as/when appropriate.

I fully expect the next 2/3 weeks to have little to no uptake (% wise), of rails 7. So we should have some time on our hands.

benny-burkhart:
Something in the dependencies might need an update -- I ended up with 1.4.0 first.
![image](https://user-images.githubusercontent.com/11530347/146694446-7137cab3-8818-48b4-8647-aa3d9be7ed46.png)

luke-hill:
Yes, as we are discussing, your PR isn't quite valid.

As a bare minimum we need to check
- New CI
- Existing features (Think we have 3/4 breakages)
- Any caveats with the rails "fake" installs.

benny-burkhart:
Hmmm I can see that CI doesn't seem to cover Rails 7... Do you know where to add that?

luke-hill:
Yes, there is already a spiked branch and PR above. If you check the OP.

tagliala:
Hi @erikaxel ,

I've investigated two of the three failures

```
cucumber features/capybara_javascript_drivers.feature:3 # Scenario: Use Browser UI backed by DB
cucumber features/capybara_javascript_drivers.feature:88 # Scenario: Use Browser UI without a DB
```

depend by the fact that Rails 7 generates a form with a `form.datetime_field :when`, so it has a modern datetimepicker and so the cucumber rails helper is not able to find date component dropdowns.

A possible fix:

```diff
diff --git a/features/capybara_javascript_drivers.feature b/features/capybara_javascript_drivers.feature
index 271639d..92f4f6b 100644
--- a/features/capybara_javascript_drivers.feature
[...]

PR #524, opened 2021-12-19
Allow Rails 7

benny-burkhart:
See issue #523


luke-hill:
Discuss in the Issue if you want to work out to fix this correctly, as mentioned we can't do this directly.

benny-burkhart:
What constructive way forward would you suggest @luke-hill ? Thanks!

luke-hill:
If you check the Issue post I have outlined the fixes required.

PR #529, opened 2022-01-18
Add ruby 3.1 & drop ruby 2.4

mgrunberg:
## Summary

- Target 2.5+ in gemspec, RuboCop config
- Fix new RuboCop offense
- Add Ruby 3.1 to github actions matrix

## Details

Rubocop requires an upgrade to 1.22.1 or higher to run on Ruby 3.1. Older versions generate the following error
```
An error occurred while Layout/BlockAlignment cop was inspecting cucumber-rails/features/support/aruba.rb:3:0. 
```
Rubocop 1.22.1 doesn't support ruby 2.4. Since ruby 2.4 EOL date is 2020-03-31 (almost two years ago) dropping support to add the latest ruby seems a good call.

The PR also downgrades ruby 2.5 version from 2.5.9 to 2.5.8. The reason is the following. Upgrading rubocop introduced a new dependency `unicode-display_width`. Bundler is unable to download it due `ArgumentError: wrong number of arguments (given 4, expected 1)`.  According to [this comment](https://github.com/rubygems/rubygems/issues/4976#issuecomment-940024665) the problem is that we're using an old version of rubygems that does not support psych 4. 
Ruby installation and bundle install is done by ruby/setup-ruby in a single step. I couldn't find a way to upgrade rubygems before running bundle install without extending the action .
Ruby 2.5.8 doesn't have this problem. Considering that ruby 2.5 EOL date is 2021-03-31 (almost a year ago), I believe it's fine to run tests against 2.5.8

## Motivation and Context

Add support to latest ruby version

## Types of changes

- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected).

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -7,21 +7,16 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        ruby: [ '2.4', '2.5', '2.6', '2.7', '3.0', '3.1' ]
+        ruby: [ '2.5', '2.6', '2.7', '3.0', '3.1' ]
         gemfile: [ 'rails_5_0', 'rails_5_1', 'rails_5_2', 'rails_6_0', 'rails_6_1', 'rails_7_0' ]
         exclude:
           # Only test latest Rails (Of each major), on actively supported Ruby Versions
           # Only test latest Ruby with latest Rails versions (Of each major)
-          # 2.4  -> Only use legacy rails versions (5.0/5.1)
           # 2.5  -> Not 5.2/6.1
           # 2.6  -> Users of this should be using Rails 5.2+
           # 2.7  -> Users of this should be using Rails 5.2+
           # 3.0  -> Rails 6.1 and 7.0
           # 3.1  -> Only 7.0
-          - { ruby: '2.4', gemfile: 'rails_5_2' }
-          - { ruby: '2.4', gemfile: 'rails_6_0' }
-          - { ruby: '2.4', gemfile: 'rails_6_1' }
-          - { ruby: '2.4', gemfile: 'rails_7_0' }
           - { ruby: '2.5', gemfile: 'rails_5_2' }
           - { ruby: '2.5', gemfile: 'rails_6_1' }
           - { ruby: '2.5', gemfile: 'rails_7_0' }
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -7,7 +7,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        ruby: [ '2.5', '2.6', '2.7', '3.0', '3.1' ]
+        ruby: [ '2.5.8', '2.6', '2.7', '3.0', '3.1' ]
         gemfile: [ 'rails_5_0', 'rails_5_1', 'rails_5_2', 'rails_6_0', 'rails_6_1', 'rails_7_0' ]
         exclude:
           # Only test latest Rails (Of each major), on actively supported Ruby Versions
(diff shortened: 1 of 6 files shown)


.github/workflows/build.yml
```diff
@@ -7,23 +7,19 @@ jobs:
...
-          # 3.0  -> Only Rails 6.1
-          - { ruby: '2.4', gemfile: 'rails_5_2' }
-          - { ruby: '2.4', gemfile: 'rails_6_0' }
-          - { ruby: '2.4', gemfile: 'rails_6_1' }
-          - { ruby: '2.4', gemfile: 'rails_7_0' }
-          - { ruby: '2.5', gemfile: 'rails_5_2' }
-          - { ruby: '2.5', gemfile: 'rails_6_1' }
-          - { ruby: '2.5', gemfile: 'rails_7_0' }
+          # 3.0  -> Rails 6.1 and 7.0
+          # 3.1  -> Only 7.0
```
mathieujobin, line 19: comments appears missplaced

mathieujobin (approved):

cucumber-rails.gemspec
```diff
@@ -2,7 +2,7 @@
 
 $LOAD_PATH.unshift File.expand_path('lib', __dir__)
 
-Gem::Specification.new do |s|
+Gem::Specification.new do |s| # rubocop:disable Gemspec/RequireMFA
```
luke-hill, line 5: can we avoid these and have a reason in the standard file.

Also this appears to be new. Should we maybe be doing this?

luke-hill (commented):
I think given the way the ruby codebase is going this makes a lot of sense. I do envisage us dropping 2.5 usage from this gem in the near future.

I would probably say we'll keep it on longer just because we do anticipate a lot more users for this to use lower ruby versions

.github/workflows/build.yml
mgrunberg: Can you elaborate or make a suggestion? I see them fine on the final file (tabs alignment and above the first exclusion item).

cucumber-rails.gemspec
mgrunberg: Moved the disabling to `rubocop.yml`. 
I'm not familiar with this setting. I'm not sure what changes are required on the release process and, sadly, I don't have time to investigate. 
I disabled the rule to defer the decision.

My ideas:
1. Defere the introduction of this setting by downgrading rubocop to 1.22.1. RequreMFA [was introduced in 1.23](https://docs.rubocop.org/rubocop/cops_gemspec.html#gemspecrequiremfa)
2. Disable RequireMFA rule and make an issue saying that we should deal with it (defere decision but have a reminder)
3. Enable RequireMFA and deal with it on next gem release

I choose number 2 but I can change it. Also open to other suggestions

luke-hill (approved):
Great work thanks. Also thanks for putting the links to the MFA thing.

On first glance I think we won't want it, but not 100%

luke-hill:
Before merging can you update changelog and readme with refs to ruby 3.1

mgrunberg:
@luke-hill updated CHANGELOG. I don't know where to put a ref on the README. I'm happy to do it if you point me in the right direction.

luke-hill:
Evidently nowhere haha. I'll merge this in AS IS.

cucumber-rails.gemspec
mathieujobin: Maybe we should deprecate ruby versions along with rails versions we choose to deprecate?

I wonder if many people runs rails 5.2 with ruby 2.7 ?

Equivalent ruby/rails version, IMO
Rails 4.2 - 2.5 works great, 2.6 needs a hack to run the tests but works great. (1.9.3 is still minimum supported)
Rails 5.x - 2.2.2 becomes minimum, 2.5/2.6 works great obviously.
Rails 6.0 - 2.5 is new minimum, we should be moving to 2.7 soonish
Rails 6.1 - 2.5 remain the minimum, 3.0 is legit candidate
Rails 7 - now requires Ruby 2.7

luke-hill: We do this with our running matrix.

PR #526, opened 2022-01-15
Add Rails 7 support

mgrunberg:
## Summary

Fixes #523

The PR was built based on issue comments.

Three tests were failing. Two because Rails 7 uses modern HTML control (input type "datetime-local") for datepickers. Capybara helper acts differently based on Rails version

The other test was failing because Destroy link is not present on index action anymore. The test changed to navigate the corresponding page based on Rails version.

PR #535, opened 2022-03-16
Add support for alternative select boxes date field helpers

### LINKED ISSUE #534: Date input support in Rails 7
The PR #526 by @mgrunberg, which was created to resolve issue #523, introduced the assumption that Rails 7 will always render date fields as `<input type="date">`. However, it's still possible for Rails 7 to render `<select>` based date inputs.

According to the [Rails Form Helpers guide](https://guides.rubyonrails.org/form_helpers.html#using-date-and-time-form-helpers):

> If you do not wish to use HTML5 date and time inputs, Rails provides alternative date and time form helpers that render plain select boxes.

There are now two officially supported form helper methods:

- `form.date_field` will output `<input type="date">` (new in Rails 7)
- `form.date_select` will output `<select>` fields (traditional Rails date input)

However when using Rails 7, only `<input type="date">` fields are supported because:

https://github.com/cucumber/cucumber-rails/blob/fa07f579ea05f6e925ff4c3ac6785214ef66e961/lib/cucumber/rails/capybara/select_dates_and_times.rb#L11-L20

### 👓 What did you see?

Cucumber no longer knows how to select dates for input fields created with `form.date_select` in Rails 7.

### ✅ What did you expect to see?

Cucumber should still be able to select dates, regardless of whether `form.date_field` or `form.date_select` was used in Rails 7.

### 📦 Which version are you using?

- cucumber-rails 2.5.0
- Rails 7

### 🔬 How could we reproduce it?

1. Install Rails 7 with cucumber-rails 2.5.0
2. Render a form with a date field created using `form.date_select`
3. Cucumber cannot select a date for this field

mgrunberg:
## Summary

Fixes #534

Rails 7 changed default scaffold datetime helper from `datetime_select` to `datetime_field`. PR #526 forces capybara helper to work only with `datetime_field`. That's fine for this gem feature specs but not for other projects.

This PR adds the ability to work with both helpers.

## Details

I'm adding another spec forcing select boxes to view helper. Both `datetime_select` and `datetime_field` have
been around since Rails 5.

### DIFF
--- a/lib/cucumber/rails/capybara/select_dates_and_times.rb
+++ b/lib/cucumber/rails/capybara/select_dates_and_times.rb
@@ -8,12 +8,13 @@ module Cucumber
         # Select a Rails date. Options hash must include from: +label+
         def select_date(date, options)
           date = Date.parse(date)
-          if ::Rails::VERSION::MAJOR >= 7
-            # Rails 7 generates date fields using input type="date". Capybara support's them
+          base_dom_id = get_base_dom_from_options(options)
+
+          # Rails 7 use HTML5 input type="date" by default. If input is not present fallback to plain select boxes alternative.
+          # It's safe to use has_css? without waiting/retry. We already know field's label is visible
+          if html5_input_field_present?(base_dom_id)
             fill_in options[:from], with: date
           else
-            base_dom_id = get_base_dom_id_from_label_tag(options[:from])
-
             find(:xpath, ".//select[@id='#{base_dom_id}_1i']").select(date.year.to_s)
             find(:xpath, ".//select[@id='#{base_dom_id}_2i']").select(I18n.l(date, format: '%B'))
             find(:xpath, ".//select[@id='#{base_dom_id}_3i']").select(date.day.to_s)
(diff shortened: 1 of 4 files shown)


lib/cucumber/rails/capybara/select_dates_and_times.rb
```diff
@@ -23,30 +24,43 @@ def select_date(date, options)
...
+          base_dom_id = get_base_dom_id_from_label_tag(options[:from])
+
+          # Rails 7 use HTML5 input type="datetime-local" by default. If input is not present fallback to plain select boxes alternative.
+          # It's safe to use has_css? without waiting/retry. We already know field's label is visible
+          if html5_input_field_present?(base_dom_id)
             fill_in options[:from], with: DateTime.parse(datetime)
           else
-            select_date(datetime, options)
-            select_time(datetime, options)
+            extended_options = options.merge(base_dom_id: base_dom_id)
```
luke-hill, line 48: Can you explain this bit here. Everything else is fine

luke-hill (commented):
Can you add a changelog entry for this bugfix. Great work though!

lib/cucumber/rails/capybara/select_dates_and_times.rb
mgrunberg: It's an optimization. I'm passing down `base_dom_id` to `select_date` and `select_time` so they don't need to find the label and extract the select/input id again.

luke-hill: Ah right I've spotted it now. It's not really extending the options, that's where I got confused. It's added the existing dom id to the options hash and then adding a guard clause to the get_base_dom_from_options method.

Although I'm not sure what I'd call it though. I think this area needs a lot of TLC and fixes but for now this will do. Thanks for explanation

luke-hill (approved):

kevindew:
@luke-hill any chance this could be released soon? we're currently pinning to a commit: https://github.com/alphagov/whitehall/pull/6483/commits/aa2b56b679aad3f37adfdf473cc93b034cdea403

luke-hill:
Probably in a couple of days if nothing else comes in

PR #581, opened 2023-10-24
v3: rubocop update

luke-hill:
### 🤔 What's changed?

Now we're about to update to v3, we should get our ducks in a row regarding rubocop. We're miles behind (Primarily because of ruby version)

Get suite more rubocop compliant, introduce new rubocop sub-gems e.t.c.

- :bank: Refactoring/debt/DX (improvement to code design, tooling, documentation etc. without changing behaviour)

### ♻️ Anything particular you want feedback on?

This will get us slightly more compliant which is good thing for v3, any concerns about styling e.t.c.


lib/cucumber/rails.rb
```diff
@@ -32,7 +32,7 @@
   MultiTest.disable_autorun
 else
   warn "WARNING: Cucumber-rails required outside of env.rb.  The rest of loading is being deferred
-  until env.rb is called. To avoid this warning, move 'gem \'cucumber-rails\', require: false'
+  until env.rb is called. To avoid this warning, move 'gem 'cucumber-rails', require: false'
```
olleolleolle, line 35: Perhaps this became a little off, nesting single quote in single quote.

luke-hill: Yep good spot. I'm not able to work on stuff much, but I think we should probably use a squiggly heredoc here.

PR #586, opened 2024-09-06
Add Rails 7.2 to test matrix

p8:
### 🏷️ What kind of change is this?

- :bank: Refactoring/debt/DX (improvement to code design, tooling, etc. without changing behaviour)

### 📋 Checklist:

- [x] I agree to respect and uphold the [Cucumber Community Code of Conduct](https://github.com/cucumber/.github/tree/main?tab=coc-ov-file)

PR #588, opened 2024-09-10
Refactor/simplify test matrix

luke-hill:
### 🤔 What's changed?

Follow up from previous work in #586 to explain the decisions. Move to an include/exclude matrix to cover old/new rubies vs standard ones

Explain the choices of what combinations to run plus simplify things a little (Removing some redundant combinations in the process).

- :book: Documentation (improvements without changing code)
- :bank: Refactoring/debt/DX (improvement to code design, tooling, etc. without changing behaviour)

### 📋 Checklist:

- [ ] I've changed the behaviour of the code
  - [ ] I have added/updated tests to cover my changes.
- [ ] Users should know about my change
  - [ ] I have added an entry to the "Unreleased" section of the [**CHANGELOG**](../blob/main/CHANGELOG.md), linking to this pull request.

----

*This text was originally generated from a [template](https://docs.github.com/en/communities/using-templates-to-encourage-useful-issues-and-pull-requests/about-issue-and-pull-request-templates), then edited by hand. [You can modify the template here.](https://github.com/cucumber/.github/edit/main/.github/PULL_REQUEST_TEMPLATE.md)*

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -34,7 +34,8 @@ jobs:
           # Capybara has an incompatibility here with rails 6. It's easier to just exclude this one run from the matrix until it's out of support
           # As such ruby 3.0 is only tested on rails 7.0 as an EXCEPTION to the above rules
           - { ruby: '3.0', gemfile: 'rails_7_0' }
-          - { ruby: '3.3', gemfile: 'rails_6_1' }
+          # Ruby 3.1+ appears to have issues with Rails 6.1 https://github.com/rails/rails/issues/46883#issuecomment-1371325906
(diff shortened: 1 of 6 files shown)


luke-hill:
cc/ @p8 / @olleolleolle for some updates here (We actually reduce the size of the matrix, but build it in a slightly diff way so you can see "how" it is built better).

luke-hill:
ping @p8 this is now merged in. I probably will update the changelog soon as this is a reasonably important change (As we're now 'officially' supporting rails 7.2 - I'll probably also cut a patch version just to signify this also.

p8:
Thanks @luke-hill !

PR #590, opened 2024-11-10
Add Rails 8 compatibility

### LINKED ISSUE #589: Rails 8 Compatibility (PR attached)
### 🤔 What's the problem you're trying to solve?

Upgrading to Rails 8.0

### ✨ What's your proposed solution?

Update gemspec

### ⛏ Have you considered any alternatives or workarounds?

_No response_

### 📚 Any additional context?

_No response_

tagliala:
In specific:
- Do not test Rails 7.0 against 3.3
- Allow 8.0 in gemspec, by relaxing the upper constraint to `< 8.1`.
  Rails uses shifted SemVer, so `8.1` is considered a Major.
- Conditionally load the sqlite3 gem based on the Rails version.
- Conditionally load selenium webdriver gem based on the Rails version.
- Simplify error page assertions to account for variations in
  apostrophes, as "We're sorry" changed to "We`re sorry" in Rails 8.

References:
- https://guides.rubyonrails.org/maintenance_policy.html#versioning
- rails/rails#53045

Close #589

---

CI: https://github.com/tagliala/cucumber-rails/actions/runs/11764599692/job/32769937431

ActiveAdmin CI against the fork (7.0, 7.1, 7.2, 8.0): https://github.com/activeadmin/activeadmin/actions/runs/11764550754/job/32769831332

---

### 🤔 What's changed?

Rails 8.0 compatibility

Unexpected downgrade to 1.4 after upgrading to Rails 8.x

- :zap: New feature (non-breaking change which adds new behaviour)

### ♻️ Anything particular you want feedback on?

n/a

### 📋 Checklist:

- [x] I've changed the behaviour of the code
  - [x] I have added/updated tests to cover my changes.
- [x] Users should know about my change
  - [x] I have added an entry to the "Unreleased" section of the [**CHANGELOG**](../blob/main/CHANGELOG.md), linking to this pull request.

----

*This text was originally generated from a [template](https://docs.github.com/en/communities/using-templates-to-encourage-useful-issues-and-pull-requests/about-issue-and-pull-request-templates), then edited by hand. [You can modify the template here.](https://github.com/cucumber/.github/edit/main/.github/PULL_REQUEST_TEMPLATE.md)*


README.md
```diff
@@ -5,7 +5,7 @@
 [![Code Climate](https://codeclimate.com/github/cucumber/cucumber-rails.svg)](https://codeclimate.com/github/cucumber/cucumber-rails)
 [![Open Source Helpers](https://www.codetriage.com/cucumber/cucumber-rails/badges/users.svg)](https://www.codetriage.com/cucumber/cucumber-rails)
 
-Cucumber-Rails brings Cucumber to Rails 5.2, 6.x and 7.x.
+Cucumber-Rails brings Cucumber to Rails 5.2, 6.x, 7.x, and 8.0.
```
tagliala, line 8: this is the "documentation change"

cucumber-rails.gemspec
```diff
@@ -26,14 +26,14 @@ Gem::Specification.new do |s|
 
   s.add_runtime_dependency('capybara', '>= 3.11', '< 4')
   s.add_runtime_dependency('cucumber', '>= 5', '< 10')
-  s.add_runtime_dependency('railties', '>= 5.2', '< 8')
```
luke-hill, line 29: Can we make this `< 9`

features/support/cucumber_rails_gem_helper.rb
```diff
@@ -29,13 +30,21 @@ def add_cucumber_rails(options)
     end
   end
 
+  def add_sqlite3_gem
```
luke-hill, line 33: Can we refactor this into the other method. In the other method, we should add selenium webdriver to be greater than 4.22 for latest rails and to not require webdrivers gem

.github/workflows/test.yml
```diff
@@ -39,15 +39,17 @@ jobs:
           - { ruby: '3.3', gemfile: 'rails_7_0' }
```
luke-hill, line 39: Can we remove testing for ruby 3.3 on rails 7.0 and 7.1 (As per documentation above)

features/allow_rescue.feature
```diff
@@ -33,7 +33,7 @@ Feature: Allow Cucumber to rescue exceptions
       end
 
       Then('I should see the public error page') do
-        expect(page).to have_content "We're sorry, but something went wrong."
```
luke-hill, line 36: Question: I get that this change is needed, but not why. Is this something to do with the way rails renders things. Is there a doc link for this I can see quickly?

luke-hill (changes requested):
Thanks for this change. Few minor tweaks and then we can get this merged in

features/allow_rescue.feature
tagliala: > but not why. 

They're just simply starting using typographic apostrophes, it's also an "issue" with validations, but they changed their mind

https://github.com/rails/rails/commit/acfa0454058c535c1352ddab91c93ca914f0805a

I will add a ref in the git commit

https://github.com/rails/rails/blame/a4e1ff72fbb24b274fce95507199c5a5f9541898/railties/lib/rails/generators/rails/app/templates/public/500.html#L7

Ref: rails/rails#53045

cucumber-rails.gemspec
tagliala: Hi sorry, I've missed this

However, Rails uses a "shifted semver version", so 8.1 should be considered as a major release 

https://guides.rubyonrails.org/maintenance_policy.html#versioning

I will change to `8.1` and a note in the commit message, but if you prefer, I can switch to 9.0

luke-hill: Thanks for the info. Let me take a read - Maybe hold off on this change for now then. Thanks for the info

features/support/cucumber_rails_gem_helper.rb
tagliala: Sorry, I may not have understood this request. I've tried what I understood

features/allow_rescue.feature
luke-hill: For now I've raised a question to rafael on the rails repo. Want to hear his thoughts before doing this.

features/support/cucumber_rails_gem_helper.rb
luke-hill: Now if you analyse the two methods `add_rails_conditional_gems` and `add_selenium_webdriver_gem` you'll see they're almost identical in branches, so they can be re-constituted back together

tagliala: > so they can be re-constituted back together

That's the first thing I've tried, but rubocop complained about complexity of the method, so I've splitted

If it is fine, I can disable rubocop on that method and merge the two methods back together

luke-hill: Ah oki, it might be better I look into tidying this up a bit more later on then. Sorry. Thanks again!

cucumber-rails.gemspec
luke-hill: Let me read this too. Sorry this is quite a lot of good info and I want to check it all out. Let me get back to you Friday/Monday (That way you're not constantly checking for updates).

mathieujobin (approved):

.github/workflows/test.yml
```diff
@@ -36,18 +36,19 @@ jobs:
           - { ruby: '3.0', gemfile: 'rails_7_0' }
           # Ruby 3.1+ has issues with Rails 6.1 https://github.com/rails/rails/issues/46883#issuecomment-1371325906
           # It (Rails 6.1.x), has been marked as a won't fix and as such it's likely this will need to just be excluded until out of support window
-          - { ruby: '3.3', gemfile: 'rails_7_0' }
           - { ruby: '3.3', gemfile: 'rails_7_1' }
```
luke-hill, line 39: Can we remove this combination too as per our rules it's just 8.all and 7.latest now

cucumber-rails.gemspec
luke-hill: Hi there. So after consideration and given the lack of response from rails. Let's set this as `< 9` and if things break we will fix them. This gem doesn't maybe get as much active updates so making the maintenance easier is better imo

luke-hill (commented):
Hi there. Can we make these 2 minor tweaks then I'll get it merged in.

luke-hill (approved):

README.md
tagliala, line 8: Since the constraint is <9, `8.0` may not play well but reflects actual compatibility

`8.0` should be changed to `8.x` after it is certain that cucumber-rails will also support 8.1

PR #598, opened 2025-09-08
Stop printing warning when `config.cache_classes` is set to false

### LINKED ISSUE #597: Remove the warning about config.cache_classes set to false?
We currently output a warning if `config.cache_classes` is set to false, we do that in this line https://github.com/cucumber/cucumber-rails/blob/b4ab57e921789f32578484824d4fd8758d89fb4c/lib/cucumber/rails.rb#L16

Right now there are two issues at least:
* In Rails 7.1 and onwards it is called `config.enable_reloading`. So new users might not understand which setting to set.
* The Spring guard does not seem to work.

I can take a stab at these issues if you would like to, but before I do that, I wanted to ask, do we really need this warning anymore? We have been running Cucumber with config_classes set to false in our test environment for several years and have never ever seen this problem. The initial warning was from 16 years ago from a bug report that no longer exists, for what I think was Rails 2. This was the earliest mention I could find in the repo: https://github.com/cucumber/cucumber-rails/blob/124b298ede48622769230fea29e04105ca899629/lib/cucumber/rails/world.rb

The only similar error we experience is if you actually change code while running a test. But that goes for rspec and minitest as well, and it does not have a similar warning afaik. I think if you change code on the fly while running tests that you reasonably expect things to go wrong.

My hunch is that with all the changes the last 16 years, in particular with Zeitwerk, that this is no longer an issue. Perhaps we could short cut this and just remove it?

erikaxel:
### 🤔 What's changed?

The initial warning was from 16 years ago from a bug report that no longer exists, for Rails 2 and is no longer relevant.
This fixed #597

Removes unnecessary noise.

- :bug: Bug fix (non-breaking change which fixes a defect)

### 📋 Checklist:

- [x]  I have added an entry to the "Unreleased" section of the [**CHANGELOG**](../blob/main/CHANGELOG.md), linking to this pull request.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 This file is intended to be modified using the [`changelog`](https://github.com/cucumber/changelog) command-line tool.
 
 ## [Unreleased]
+### Removed
+- Stop printing a warning about `config.cache_classes` being set to `false`
 
 ## [4.0.0] - 2025-09-01
 ### Changed
--- a/lib/cucumber/rails.rb
+++ b/lib/cucumber/rails.rb
@@ -12,16 +12,6 @@ if called_from_env_rb
   require File.expand_path("#{ENV.fetch('RAILS_ROOT')}/config/environment")
   require 'cucumber/rails/action_dispatch'
   require 'rails/test_help'
-
-  unless Rails.application.config.cache_classes || defined?(Spring)
-    warn <<~MESSAGE
-      WARNING: You have set Rails' config.cache_classes to false (Spring needs cache_classes set to false).
-      This is known to cause problems with database transactions.
-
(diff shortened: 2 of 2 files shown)


CHANGELOG.md
```diff
@@ -8,6 +8,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 This file is intended to be modified using the [`changelog`](https://github.com/cucumber/changelog) command-line tool.
 
 ## [Unreleased]
+- Stop printing a warning about `config.cache_classes` being set to `false`
```
luke-hill, line 11: Because of the way our changelog parser / automation works these all need a subheading. This should be `### Removed` (See in the v4 / v3 releases how they're structured.

luke-hill (approved):
For now this will live in an approved PR. I need to have a think about versioning in the v4 lifecycle.

You don't need to do anything more.

PR #612, opened 2026-06-20
Allow Cucumber 11

JulienItard:
## 🤔 What's changed?

Bumps the `cucumber` runtime dependency constraint in the gemspec from `< 11` to `< 12`, so cucumber-rails can be used with the Cucumber 11.x line (currently 11.1.0).

```diff
- s.add_runtime_dependency('cucumber', '>= 7', '< 11')
+ s.add_runtime_dependency('cucumber', '>= 7', '< 12')
```

The legacy Rails 6.1 / 7.0 / 7.1 appraisal matrices keep their existing older-cucumber pins; the `rails_8_0` appraisal inherits the gemspec range and resolves to Cucumber 11. As per CONTRIBUTING, the CHANGELOG/version are left untouched.

## ⚡️ What's your motivation?

Cucumber 11 has been released, but downstream apps can't adopt it while cucumber-rails caps the dependency at `< 11`. This unblocks them.

## 🏷️ What kind of change is this?

- [x] Refactor (improvement to existing code without functional change)

## ♻️ Anything particular you want feedback on?

I validated locally against `gemfiles/rails_8_0.gemfile` with the dependency resolving to **cucumber 11.1.0**:

- **RSpec:** 22 examples, 0 failures.
- **Cucumber features (non-`@javascript`):** identical pass/fail results to a baseline run pinned at `cucumber ~> 10.0` — same container, same filter, only the cucumber version differing, so this bump introduces no regression.

I did not loosen the `< 11` / `< 10` pins in the older-Rails appraisal gemfiles, on the assumption those are intentional for the legacy CI matrix — happy to adjust if you'd prefer them updated too.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 This file is intended to be modified using the [`changelog`](https://github.com/cucumber/changelog) command-line tool.
 
 ## [Unreleased]
+### Changed
+- Allow Cucumber 11.x by bumping the runtime dependency cap to `< 12` [#612](https://github.com/cucumber/cucumber-rails/pull/612)
 
 ## [4.0.1] - 2026-03-10
 ### Removed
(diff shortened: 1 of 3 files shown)


luke-hill:
Can you please add a changelog entry.

I'm aware that file says not to. That's a mistake. If you can delete that mini bit too that would be helpful :sweat_smile:

JulienItard:
@luke-hill Done ! I added at first then saw the mention and removed it. Back again !

luke-hill:
FYI this is released as cucumber-rails 4.1

